PR #66, opened 2015-04-02
Fix up mount point directory variable name

cpschult:
Fix up mount point directory variable name in volume attach routines in cb_common.sh.


cpschult:
Only wanted one of the change sets to come across.  Retrying ....

PR #72, opened 2015-05-09
fixes for hadoop scripts. adding parameters in virtual application th…

salmanbaset:
fixes for hadoop scripts. adding parameters in virtual application that correspond to SPEC hadoop image. Added commented parameters in cb_hadoop_common.sh for setting map and reduce tasks


ibmcb:
Salman and I (Marcio) sat together and coded a better, more robust solution for it.

PR #76, opened 2015-08-18
Translate _dash_ to '-' in OSK access URL value.

cpschult:
Allow hostnames in ACCESS URL to contain '-'.  ( A valid character in hostnames.)


cpschult:
Don't delete YCSB result file. Useful for debugging.

cpschult:
Adding clock synchronization patch.  This code runs when a new VM is being configured (attached) to cbtool.  ntpdc seems to be installed along with ntp package on RHEL and ubuntu.  Is it installed on your VMs Marcio?

PR #84, opened 2015-11-02
DigitalOcean support, Tested with YCSB/Cassandra and Nullworkload

hinesmr:
Many things were done in this commit:

1. Starting from the VMware adapter was a bad idea --- it was basically shit, but I was in too deep to pull out, so this commit completes the libcloud-based DigitalOcean adapter which is *actually* fully tested against the cloud. Only vmcreate/vmdestroy is tested --- not the other functions like vmcatpure, etc, as they are not required by SPEC right now.

2. VPN support was *still* broken, and I have fixed it (again). It appears that the decision has been made to make the individual cloud adapters exclusively responsible for the own VPN management, so our adapter uses cloud-init exclusively to ensure that all VMs are part of the VPN and does nothing to help anybody else's cloud adapter.

3. YSCB/Cassandra failed to work initially because Cassandra disabled the JMX port (7199) to listen on the localhost ---- so the cassandra scripts needed to be updated to force cassandra to listen on a usable IP address.

4. There are also bug fixes to gmetad-python --- please accept that pull request as well.


ibmcb:
Just merged the commit as is, thank you.

The only thing broken for me with this commit was the VPN support (it was working before, not working now). I will have a commit of my own very soon with some fixes.

hinesmr:
The re-worked VPN support you committed never worked because the tool was
not using VPN ip addresses for bootstrapping the VMs, nor was mongodb or
gmetad or anything else using VPN ip addresses. My patches fix all that
stuff.

The tool still uses normal internal IP addresses for intra-workload
communication, of course.

hinesmr:
Here is how I use the VPN, which allows all aspects of the tool to function
over the VPN not including intra-workload traffic:

[VM_DEFAULTS : CLOUDOPTION_MYDIGITALOCEAN ]
USE_VPN_IP = $True
USERDATA = $True    # This is implemented exlusively by the digitalocean
driver and must be copied to other drivers
                                    # which also preserves the original
userdata requested from lib/auxiliary/data_ops.py

[VPN : CLOUDOPTION_MYDIGITALOCEAN ]
SERVER_IP = 159.203.103.89         # public VPN server using the
server.conf file from cbtool/configs/generated
SERVER_BOOTSTRAP = 10.9.0.6   # The VPN-assigned IP address of the
orchestration server
NETWORK = 10.9.0.0                       # User-chosen
SERVER_PORT = 3001                    # User-chosen

ibmcb:
Thanks for the clarification. I was a puzzled by the categorical affirmation "The re-worked VPN support you committed never worked", since I have been using this without any issue for quite some time :-) But now I realize source of confusion. There are actually two use cases for VPN:

1) In clouds such as OpenStack (as far as I know, this is true for CloudStack and VCloud, but not sure enough to state with certainty), the VMs on private networks can actually properly connect to Object, Metric and Log stores outside of the cloud through routing (in OpenStack this is done through neutron L3 router(s). In this scenario, the orchestrator only need to use the OpenVPN-assigned IP address once, during the initial deployment, to run the deployment scripts (yes, I know, this is also used for airesize).

2) In other clouds, we might want (or maybe need?) the VMs to reach the Object, Metric and Log stores outside of the cloud through the VPN server. In this scenario, I imagine that the VPN server could be even outside the orchestrator, probably on a multi-homed node that can be reached from the inside and outside the cloud, very much like a "jump box".

Important: In both cases, the *Orchestrator-initiated traffic* (i.e., "south-north") has always to go through the OpenVPN-assigned IP addresses (which was the case for the OpenStack cloud adapter already). Also, as you correctly pointed out, private IP addresses are used for intra-workload communication (i.e., east-west), no matter what.

Being that the case, my next commit will include the following parameters:

a) USERDATA=$True 

Will cause the generation of cloud-init script that simply pulls (rsync, etc.) a copy of the code from the orchestrator and also publishes a message indicating the end of the boot. This is what is done today and it is cloud-agnostic.

b) USE_VPN_IP = $True

The cloud-init script will *FIRST* execute the OpenVPN client to connect to the VPN server, and then execute the same steps as a). However all *VM initiated traffic* to the Stores (in the Orchestrator) is done through "normal", "outside of the cloud" IP addresses. 

c) ONLY_VPN = $True

Same as b), but now all *VM initiated traffic* to the Stores (in the Orchestrator) is through OpenVPN-assigned IP addresses.

As you can see, a lot of the changes affect directly the userdata generation. For now, I am keeping the userdata generation as a cloud-agnostic method. If you believe that it might result in too many breakages, I am quite open to have it moved to cloud-specific portion of the tool. It will result in some code duplication (not much), but will certainly make easier and safer for other to contribute other cloud adapters.

hinesmr:
Thanks for the clarification. Responses:

1) and 2) Yes. The re-worked support was singly focused on openstack and
did not support all clouds.

Regarding a redefinition of a) and b), This is too openstack-specific. I do
not believe users that are not familiar with neutron will understand what
this means and makes VPN support too complicated.

As I have modified it now: The effect of c) "ONLY_VPN" is already finished
via a combination of both USERDATA and USE_VPN_IP, the way I have patched
it, and solves the implementation across all clouds that support user-data,
including Amazon, Openstack, and DigitalOcean (not sure about VMware). I
believe a) and b) should be subsets of c) and are all solved by c).

The problem as that, until now, cloudbench was not using userdata correctly
as a mechanism of cloud-init and cloud-config. A "strict" use of cloud-init
(including on Amazon, where it originated), requires cloud-config syntax,
and that was not being done.

Take a second to examine my use of cloud-init in do_cloud_ops.py, which
does make correct use of cloud-config syntax, which is supported by both
RHEL and Ubuntu systems today with cloud-init properly installed.

When you generate userdata in data_ops.py, the DigitalOcean adapter than
takes that as a single file and wraps it in cloud-config syntax along with
the necessary cloud-specific VPN commands needed to connect the VM to the
VPN server *before* bootstraping the VM itself. It works both if the users
chooses to use the VPN or not use the VPN. This can only be done through
full use of cloud-config and cannot be done simply by rsyncing the VPN
configuration files, as rsync itself wouldn't work without a fully
connected VPN on all sides.

In fact, I grew to like the way you re-worked VPN support, because it puts
the burden use of the VPN to the individual cloud providers. I believe this
is a good thing because cloud users must be aware that cloud-init needs to
be running in their virtual machines at boot: I believe the function I've
implemented for cloud-config in do_cloud_ops.py should be generalized to
lib/clouds/shared_functions.py, and would be willing to do that if
openstack users are aware that they need to have cloud-init installed
inside the VMs correctly.

- Michael

ibmcb:
Thank you again for the thought-through considerations.

You're correct, the implementation was tested and used on OpenStack only. I fear I gave you the wrong impression regarding OpenStack. While it is true that I have a much better knowledge of the exact mechanism employed by OpenStack (i.e., neutron L3 routers) to implement north-south traffic, I believe a similar case can be made for any private/local cloud (i.e., CloudStack, VCloud, our own pure KVM and pretty soon, pure Swarm/Fleet ).

Given that the south-north traffic in a private/local cloud is not metered, users (and benchmarkers) will prefer to have "telemetry" (i.e., Object, Metric and Log Store) flowing through whatever mechanism (typically virtual routers of some sort) is already provided by the cloud.

On public cloud, given that north-south is typically metered, your solution (i.e., "telemetry" over VPN) is almost certainly the only viable alternative. Seeing your implementation and reading your comments, I am under the impression that you do believe that actual OpenVPN server is/will be/should be in a node separate from the orchestrator. I am correct in my belief?

Anyway, I believe we are actually in agreement regarding the need for (or at least the conceivability of) 3 configurations: 

a - incoming/outgoing "telemetry" traffic over "native" network
b - incoming "telemetry" traffic over VPN network, outgoing "telemetry" traffic over "native" network
c - incoming/outgoing "telemetry" traffic over VPN network

The point where I (friendly) disagree is the use of just USERDATA and USE_VPN_IP parameters to control it. In my view (I fully accept that I can be wrong on that), USE_VPN_IP = $True will *always require USERDATA = $True* no matter what. We could envision a scenario where the images are burned it with a OpenVPN configuration file that already points to the correct server (this making USERDATA = $False possible), but I think you will agree that this not quite practical. Being that as it is, we are left with only one "bit" (USE_VPN_IP) which can only represent 2 states: a,b. That is why I don't see any way around introducing the second "bit" ONLY_VPN.

Regarding the USERDATA generation. The current contents are good enough for cloud-init execution in OpenStack (I tested only with the "config_drive" option in nova), but I am certainly open to have that generated in a more canonical manner. Yes, I saw your implementation and believe that it would be very wise do have your data generation as a method in shared_ops.

Before seeing your answer, I had already made some changes to accommodate both the 3 configurations and multiple clouds clouds. I will push what I have so far, but please do not consider this the end of the discussion, just one more iteration towards more robust VPN solution. My apologies in advance if this breaks again the VPN support in Digital Ocean in any form (I don't think it will, but do not have a good way to test it). Let's continue to refine it.

hinesmr:
I think we're converging converging on an agreement. Thank you. Responses:

>> On public cloud, given that north-south is typically metered, your
solution (i.e., "telemetry" over VPN) is almost
>> certainly the only viable alternative. Seeing your implementation and
reading your comments, I am under the
>> impression that you do believe that actual OpenVPN server is/will
be/should be in a node separate from the
>> orchestrator. I am correct in my belief?

Yes, that's correct. The use cases are listed here. Please add additional
OpenStack-specific use cases for which that gathering of non-metered
telemetry is possible, after which we should have covered all the use
cases:
https://github.com/ibmcb/cbtool/wiki/HOWTO:-Use-VPN-support-with-your-benchmarks

>>The point where I (friendly) disagree is the use of just USERDATA and
USE_VPN_IP parameters to control it.

I have no problem altering parameters or re-working the VPN implementation
to converge on a solution for everybody.

How about we collapse/rename/reorganize the two parameters into the
metered/unmeterd versions that you mentioned, something like:

PRIVATE_VPN = $True # indicates to CloudBench that a white-box cloud uses a
non-metered VPN link by which the user's VPN configuration files can be
rsync'd directly without the need of cloud-init installed manually by the
user
PUBLIC_VPN = $True # indicates to CloudBench that a black-box cloud uses a
metered VPN link by which users have no insight into the underlying L2/L3
infrastructure of the cloud and are limited only by the public/private IP
addresses assigned by the cloud.

Userdata is not strictly necessary for PRIVATE_VPN, because if you have the
ability in a white-box enviroment like a private OpenStack to rsync the VPN
configuration file (As was done before), then you should also have the
ability to manually start the VPN server without the use of any userdata or
cloud-init installed either.

All the names are open for discussion, of course --- but the fewer the
parameters the better, as it makes it easier for users of multiple clouds
to understand the terminology.

We can figure out lib/clouds/shared_functions.py later after we've first
agreed on the best cloud-agnostic solution to use.

ibmcb:
Great, we are almost there, I believe.

Important point: I fully agree with PRIVATE_VPN and PUBLIC_VPN. These names have the added benefit of being a lot more descriptive than what I initially came up with. Do you want me to change these, or do you want to make another pass?

A couple of non-consequential comments.

#1) If (PRIVATE_VPN == $True OR PUBLIC_VPN == $True) and VPN_START_SERVER == $False, shouldn't we (CB) start a VPN client for the user, automatically setting the VPN_SERVER_BOOSTRAP with whatever is handed to us by the external OpenVPN?

#2) Conversely, if (PRIVATE_VPN == $True OR PUBLIC_VPN == $True) and VPN_START_SERVER == $True, and we (CB) already start a VPN server, shouldn't we also automatically set the VPN_SERVER_BOOSTRAP?

hinesmr:
For #1) In your last implementation there was no SERVER_BOOTSTRAP at all
------- did you actually need it?

Setting SERVER_BOOTSTRAP should be done by the VPN client itself: For
example, the way I'm doing it in the do_cloud_ops.py implementation: Upon a
final successful connection by OpenVPN: OpenVPN has the ability to
"callback" a script with all the new IP address parameters of the
connection. That's how it works with cloud init/cloud-config.

The client side (where CB is running) should, in the end, do the same
thing. It works by specifying the "up" command in the OpenVPN configuration
file.

This is the most reliable way of doing it because if you reboot your
machine, the client VPN IP address (and thus SERVER_BOOSTRAP) will change
to a new IP address issued by the VPN server, and you will need a callback
from the client itself anyway to keep that value updated.

ibmcb:
Ok, I must admit that I did not have contemplated the need to re-update the OpenVPN-assigned IP address upon an eventual reboot. Please go ahead and re-work it as needed. 

Three last quick comments:

#1) if PRIVATE_VPN = $True, then there is no need to use SERVER_BOOTSTRAP parameter

#2) Just for clarity, here is what was done in my last commit. 
a) Generate a user-data that already included the generated client openvpn.conf file "inline" (this manner, there is no need to rsync - push or pull - any OpenVPN-related file to the client)
b) Use this file to connect back to the OpenVPN server
c) Get the OpenVPN-assigned IP address of tun0, and write it to the PENDING version of the VM object. 
d) ... Whatever else is normally done in the user-data (i.e., generate the cb_os_parameters.txt, pull, via rsync, a copy of code, etc.) ...

#3) In the re-working that you're describing, which involves a callback script (did not know this feature of OpenVPN) it is my understanding that the client (i.e., the VM itself) will have to update both the "regular" and the PENDING version of the VM object with the OpenVPN-assigned IP, correct? The former is required in case the IP changes upon reboot, while the latter is required during the initial deployment. Is this what you were thinking too?

hinesmr:
You'll have to excuse me, but I'll be going off the grid for several weeks.
I know it's bad timing, but my son was just born today in China, and I'm
here on paternity leave starting right now with my wife Chen.

My short wrap-up answer is, yes, the existing OpenVPN callback mechanism
solves a lot of these issues --- I have a working example in that last
patchset that was sent to you which is used by cloud init by uploading the
file cbtool/util/openvpn/client_connected.sh, which is used in
do_cloud_ops.py.

My colleague, Darrin Edin, will do his best to synchronize with SPEC during
my absence, but for the moment, I've submitted a sample (example only) SPEC
baseline + elasticity runs to Salman already by Skype as proof that
everything as working.

Was missing on DigitalOcean are the public images. I think it's OK to
release those in the future ---- it need not impede the progress made by
the committee.

Thanks again for your patience --- and good luck with the release.

- Michael

jtaleric:
Congrats on the baby!! :D

On Mon, Nov 9, 2015 at 4:09 AM, Michael R. Hines <notifications@github.com>
wrote:

> You'll have to excuse me, but I'll be going off the grid for several weeks.
> I know it's bad timing, but my son was just born today in China, and I'm
> here on paternity leave starting right now with my wife Chen.
>
> My short wrap-up answer is, yes, the existing OpenVPN callback mechanism
> solves a lot of these issues --- I have a working example in that last
> patchset that was sent to you which is used by cloud init by uploading the
> file cbtool/util/openvpn/client_connected.sh, which is used in
> do_cloud_ops.py.
>
> My colleague, Darrin Edin, will do his best to synchronize with SPEC during
> my absence, but for the moment, I've submitted a sample (example only) SPEC
> baseline + elasticity runs to Salman already by Skype as proof that
> everything as working.
>
> Was missing on DigitalOcean are the public images. I think it's OK to
> release those in the future ---- it need not impede the progress made by
> the committee.
>
> Thanks again for your patience --- and good luck with the release.
>
> - Michael
>
> On Sun, Nov 8, 2015 at 12:24 AM, ibmcb <notifications@github.com> wrote:
>
> > Ok, I must admit that I did not have contemplated the need to re-update
> > the OpenVPN-assigned IP address upon an eventual reboot. Please go ahead
> > and re-work it as needed.
> >
> > Three last quick comments:
> >
> > #1 <https://github.com/ibmcb/cbtool/pull/1>) if PRIVATE_VPN = $True,
> then
> > there is no need to use SERVER_BOOTSTRAP parameter
> >
> > #2 <https://github.com/ibmcb/cbtool/pull/2>) Just for clarity, here is
> > what was done in my last commit.
> > a) Generate a user-data that already included the generated client
> > openvpn.conf file "inline" (this manner, there is no need to rsync - push
> > or pull - any OpenVPN-related file to the client)
> > b) Use this file to connect back to the OpenVPN server
> > c) Get the OpenVPN-assigned IP address of tun0, and write it to the
> > PENDING version of the VM object.
> > d) ... Whatever else is normally done in the user-data (i.e., generate
> the
> > cb_os_parameters.txt, pull, via rsync, a copy of code, etc.) ...
> >
> > #3 <https://github.com/ibmcb/cbtool/pull/3>) In the re-working that
> > you're describing, which involves a callback script (did not know this
> > feature of OpenVPN) it is my understanding that the client (i.e., the VM
> > itself) will have to update both the "regular" and the PENDING version of
> > the VM object with the OpenVPN-assigned IP, correct? The former is
> required
> > in case the IP changes upon reboot, while the latter is required during
> the
> > initial deployment. Is this what you were thinking too?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ibmcb/cbtool/pull/84#issuecomment-154719779>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ibmcb/cbtool/pull/84#issuecomment-155001999>.
>



-- 
Joe Talerico
Mobile : 919 760 3476

ibmcb:
Hey Michael, congratulations to you and Chen on the baby! 

Worry not, what I have in place is perfectly functional for our current needs on OpenStack (VPN-wise). I will work with Darrin, as needed, to incorporate all changes that we discussed so far.

PR #87, opened 2015-11-16
A few last-minute bugs for DigitalOcean

hinesmr:
A few last-minute bugs for DigitalOcean
    
    1. When running the SPEC harness, we can only allow one VM per AI
       to be created at the same time. More than that, the DigitalOcean
       scheduler appears to add too much delay.
    
       NOTE: DigitalOcean has a 10 or 20 VM limit by default anyway, so
       without further contact with the company through a support ticket
       users will not be able to complete the elasticity run anyway.
       The baseline run will be fine, but the elasticity run will not work.
    
    2. One VMC related bug and setting the default VMC to New York.

### DIFF
--- a/configs/cloud_definitions.txt
+++ b/configs/cloud_definitions.txt
@@ -103,11 +103,8 @@ DO_KEY_NAME = 1520229                                   # Comma-separated list o
 # Your DO ssh key IDs can be found like this: curl -H "Authorization: Bearer YOUR_ACCESS_TOKEN" -X GET "https://api.digitalocean.com/v2/account/keys"
 
 [VM_TEMPLATES : CLOUDOPTION_MYDIGITALOCEAN ]
-TINYVM = size:512mb, imageids:1, imageid1:13991071
+TINYVM = size:512mb, imageids:1, imageid1:foo
 # size options are: 512mb, 1gb, 2gb, etc....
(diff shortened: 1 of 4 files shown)


ibmcb:
Hello Michael,

Just to confirm. Do you want to leave the section [VM_TEMPLATES : CLOUDOPTION_MYDIGITALOCEAN ] as part of the generic cloud configuration file? 
We do not have a section like that for any other cloud, and I just wanted to double-check to make sure that it wasn't just some leftover from some test.

hinesmr:
Sorry for the late reply ---- I'm up all night dealing with my son on a 12-hour delay =).

Yes, the changes to cloud_definitions.txt are intentional: While I was performing testing, the DigitalOcean public cloud can't handle more than 2 simultaneous provisioning requests per user, which limits the SPEC runs to 1 virtual machine per AI at any given time, so I want the defaults to reflect that in case a user wants to run a long-running experiment against DigitalOcean.

Additionally, new user accounts are limited to about 5 or 10 VMs and users need to send a support ticket to get their limit increased.

Note: In any particular DigitalOcean data center, there may be hundreds of events happening at the same time, but the per-user handling (as a public cloud) do not lend themselves well to intra-user parallelism, all though we do have lots of inter-user parallelism.

Hope that was clear....

ibmcb:
My apologies for the delayed acceptance. 

Yes, it was very clear, thank you very much.

PR #103, opened 2016-05-04
Add Dockerfile to run cbtool container on centos image

psuriset:


hinesmr:
:+1: Thank you for this.

PR #107, opened 2016-05-18
Remove Dockerfile as we have two seperate files for centos & ubuntu

psuriset:


ibmcb:
Already removed the old file. Thank you Pradeep!

PR #109, opened 2016-06-17
SPECCloud-requested changes, split in incremental commits for easier review

ibmcb:


lib/operations/active_operations.py
```diff
@@ -419,11 +417,15 @@ def start_vpnserver(self, cld_attr_lst) :
             vpn_client_config += '/' + cld_attr_lst["cloud_name"] + "_client-cb-openvpn.conf"
             cld_attr_lst["vm_defaults"]["vpn_config_file"] = vpn_client_config
 
-            if str(cld_attr_lst["vm_defaults"]["use_vpn_ip"]).lower() == "false" :
-                return
-
+            if cld_attr_lst["vm_defaults"]["use_vpn_ip"] and not cld_attr_lst["vpn"]["start_server"] :
```
hinesmr: Why was this re-added? Can this section be deleted, please? I need to be able to use START_SERVER = $False in my configuration file. If you need START_SERVER to be set to True, can you please do it in your private configuration file?

ibmcb:
Certainly, makes perfect sense. Will remove again.

PR #114, opened 2016-07-05
Improvements to vCloud Director driver

bowers:
Allow launching images (vApps in vCloud Director terminology) from the vCloud Director catalog, as well as cloning instantiated images.  Move vCloud Director API credentials (username/password) into single CREDENTIALS parameter.  Drop some unnecessary parameters in template file.  Change default behavior on creating new VM to poll_cloud for success rather than assuming instance is launched.  Improve VM termination code to ignore "unlaunched" instances.  Various other minor improvements. Also fixex Issue 82, in which password in vCloud Director template conflicting with password used for mongodb.


hinesmr:
A great test for this would be to run SPEC against this. Care to give it a wirl? We're all in a meeting both today and tomorrow all today. You're welcome to join.

hinesmr:
@maugustosilva @bowers Was able to get SPEC to run against this PR ..... @bowers comments?

bowers:
Yes, I ran the SPEC benchmark against this, and it worked.   (I had to increase the AI arrival rate of the SPEC benchmark because my testbed has very slow provisioning times, so I couldn't do compliant runs, but it did work.)

PR #119, opened 2016-07-15
Fix for submitter always at 100% CPU

hinesmr:
The calculation in the submitter is always negative, and thus always causes the submitter to sleep for (5 / 10) = 0.5 seconds, which always appears to be spinning at 100%. Two submitters doing this thus takes up 2 whole cores. 

Using the absolute value instead fixes the problem ---- but I'm not entirely sure if that's what you were going for @maugustosilva.

The second commit is for DigitalOcean --- a printing error.


ibmcb:
Hey Michael,

Yes, you are correct, I should have used absolute values. Thank you.

PR #120, opened 2016-07-20
fix wrong DigitalOcean provisioning time

hinesmr:
We were calculating "request_originated" incorrectly. Fix that.

The same thing probably goes for alibaba and VMware, as those things were carried over from the older vmware implementation, but those are not fixed here.

### DIFF
--- a/lib/auxiliary/gui.py
+++ b/lib/auxiliary/gui.py
@@ -66,6 +66,7 @@ class Dashboard () :
         self.time_vars = time_vars 
         self.base_uri = base_uri
         self.start_time = int(self.time_vars["start_time"])
+        self.pid = "none"
         self.processid = "none"
         self.cn = cloud_name
         self.msattrs = msattrs
--- a/lib/clouds/do_cloud_ops.py
+++ b/lib/clouds/do_cloud_ops.py
@@ -491,6 +491,8 @@ class DoCmds(CommonCloudFunctions) :
 
             cbdebug("Launching new Droplet with hostname " + obj_attr_list["cloud_vm_name"], True)
 
+            obj_attr_list["mgt_001_provisioning_request_originated"] = int(time())
+
             _reservation = catalogs.digitalocean[credential_pair].create_node(
                 image = image,
                 name = obj_attr_list["cloud_vm_name"],
@@ -576,13 +578,7 @@ class DoCmds(CommonCloudFunctions) :
             _status = 100
             _fmsg = "An error has occurred, but no error message was captured"
 
-            _time_mark_drs = int(time())
             _wait = int(obj_attr_list["update_frequency"])
-
-            if "mgt_901_deprovisioning_request_originated" not in obj_attr_list :
-                obj_attr_list["mgt_901_deprovisioning_request_originated"] = _time_mark_drs
-
-            obj_attr_list["mgt_902_deprovisioning_request_sent"] = _time_mark_drs - int(obj_attr_list["mgt_901_deprovisioning_request_originated"])
 
             cbdebug("Last known state: " + str(obj_attr_list["last_known_state"]))
 
@@ -597,6 +593,7 @@ class DoCmds(CommonCloudFunctions) :
             _msg += "...."
             cbdebug(_msg, True)
 
+            firsttime = True
             while True :
                 _errmsg = "get_vm_instance"
                 cbdebug("Getting instance...")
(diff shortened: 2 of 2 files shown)


ibmcb:
Hello Michael,

For every other cloud, "mgt_001_provisioning_request_originated" is defined outside the cloud-specific adapter code. The reason for this is the use of "mgt_002_provisioning_request_sent", calculated just after all "preparation" tasks (usually the instantiation of a new cloud-specifc client connection) are complete, to estimate the "cloud-specific client/connection overhead" (subtracted from mgt_001, of course). Since we never discussed it formally, I would point to the OpenStack adapter code, currently on the "experimental" branch, as an example of what I had in mind: https://github.com/ibmcb/cbtool/blob/experimental/lib/clouds/osk_cloud_ops.py#L2629

Is there a specific reason why you would like to change that? I am not a priori opposed to it, I just wanted to understand the need for it better. Please note that I am also open to the introduction of a new mgt_X measurement, just to account for the time spent on the "generic" portion of the code versus the amount spent on the "cloud-specific" portion, *right before* the issuing of the actual instance creation.

hinesmr:
@maugustosilva Yes, the current positioning of "request_originated" is a significant problem for libcloud-based adapters like DigitalOcean.

The reason is that, due to the object model of libcloud, all parameters to libcloud functions are Class-objects --- libcloud-defined objects the must be looked up before every API call by communicating with the cloud.

Looking up these objects (locations, images, networks, etc...) takes many 10s of seconds, despite the fact that these objects NEVER change during a benchmark.

When a new AI create daemon is forked, these object lookups looks happen from scratch, since it's a brand new process. When this happens on the command-line, this is non-issue because the adapter knows how to cache these objects before talking to the cloud to create the actual virtual machine. But, when a new daemon is forked, these objects must be re-populated from scratch by re-communicating with the cloud to look them up before the VM can actually be created.

As a result of this artificial library-imposed delay, the time between "request_originated" and "request_sent" is, unfortunately being accounted for in the SPEC harness, and as a result this additional delay (upwards of 30 seconds or more) is incorrectly contributing to VM instance provisioning time.

That's the primary reason for the change.

I certainly have no problem with introducing a 3rd parameter in between the two parameters as you mentioned above, but we would also need permission from SPEC to modify or be aware of this parameter, and that seemed problematic.

I'm totally open to suggestion though ---- but that's the cause of the problem: SPEC is accounting for a delay that doesn't really have anything to do with the actual provisioning of the VM, and that needs to be fixed.

A 3rd option might be to modify the location of "request_originated". If "request_originated" could be populated AFTER aidefine(), then we (the adapter itself) could perform the object lookups inside of aidefine() before the timestamp was taken ---- that way the timestamp wouldn't have to renamed at all and the delay between "request_originated" and "request_sent" would not be artificially inflated. That seems like it might be a better compromise......

Does that make sense?

hinesmr:
Here is an example from the dashboard of the invalid delay before the patch:

before

 And after the patch:

after

See the 33-seconds instead of 1-second? That's the problem. It's due to the way libcloud works above.

hinesmr:
@maugustosilva aidefine() to the rescue! I have 1-liner solution to this using aidefine(). No need to override "request_originated" anymore. 

Yayyy. I'm using the same PR for this fix. Please review.

ibmcb:
Two comments:

1) The use of aidefine for this purpose is brilliant. Thank you very much!
2) The problem that you pointed out on the previous message (the 33 second versus 1 second), will still exist for the VMs, right? While this is not something that I am in hurry to solve, you're completely right... at some point we will have to either introduce an intermediate mgt (after clearing with SPEC), or stop using "mgt_002_provisioning_request_sent" as proxy measurement for the "core CB overhead" (at least for individual VMs).

Thanks again. Will review and merge soon.

hinesmr:
@maugustosilva To answer question 2) Using aidefine() solves this because since the libcloud objects get cached inside of aidefine(), the timestamps for the individual VMs have not yet been taken ----- that way, the timestamps look normal. Meaning the delay between "request_originated" and "request_sent" is very small, since we've already cached the libcloud objects directly inside the same process / driver.

So, we're good to go ---- no need to change the rest of the codebase at this time.

ibmcb:
oh, sorry... I was completely inarticulate in formulating the question :-) 

What I meant is, it does not solve the problem of standalone "vamattach" API calls, which are - I readily concede - extremely rare (mostly for debugging). This is certainly a non-urgent problem.

hinesmr:
Oh, yes, yes --- that's correct. That doesn't bother me because individual vmattach API calls are not used in SPEC at all ----- and in general, if someone is using individual vmattach calls, they are likely using the command line anyway and not the API. 

So, for all common cases, the problem is solved :)

salmanbaset:
this PR will need to be discussed in SPEC Cloud Wednesday meeting. Pictures will be helpful.

hinesmr:
@salmanbaset @maugustosilva Was this picked up yet?

PR #122, opened 2016-07-27
Added UnixBench as a workload for CBTool

chriskirkland:
Adds functionality for running UnixBench as a CBTool AI workload.  Supports the official UnixBench Index test (or any subset therein).


hinesmr:
We'll get this in soon (I also have an additional workload to add), but we're in a freeze right now for a minor release of SPEC Cloud 2016.

We'll get back to this soon.

scripts/unixbench/cb_unixbench.sh
```diff
@@ -0,0 +1,202 @@
...
+then
+    syslog_netcat "Usage: cb_unixbench.sh <load profile> <load level> <load duration> <load_id>"
+    exit 1
+fi
+
+UNIXBENCH_DIR="~/byte-unixbench/UnixBench"
+eval UNIXBENCH_DIR=${UNIXBENCH_DIR}
+
+UNIXBENCH_IP=`get_ips_from_role unixbench`
+source ~/cb_barrier.sh start
```
hinesmr: Remove this barrier line, please. It's a duplicate. I've a PR coming that also fixes the duplicates in other workloads.

scripts/unixbench/dependencies.txt
```diff
@@ -0,0 +1,26 @@
...
+### END - Dependency-specific installation method ###
+
+### START - Tests ###
+### END - Tests ###
+
+### START - Dependency versions ###
+unixbench-ver = ANY
+### END - Dependency versions ###
+
+### START - Dependency URLs ###
```
hinesmr: This is not complete. It should include a specific set of accurately-crafted commands that any user can copy-and-paste to make their local environment benchmark-ready.

Even if you're using "man" option, the URLs still need to be there for user references purposes.

scripts/unixbench/virtual_application.txt
```diff
@@ -0,0 +1,29 @@
+# Parameters for this Virtual Application (Application Instance - AI) type should
+# be set on YOUR private configuration configuration file, including the ones 
+# commented.
+
+[AI_TEMPLATES : UNIXBENCH]
```
hinesmr: Please also add a section to `configs/templates/PUBLIC_application_submitters.txt`. I know we don't have the full set of applications in there, but let's make it a habit anyway.

Example PR also coming for that.

hinesmr:
@chriskirkland PR #125 is a more complete example. Skim through that one at your convenience.

chriskirkland:
@hinesmr After review in the sample PR and your comments, I've made some modifications/additions.  Please review at your earliest convenience.  Thanks!

scripts/unixbench/dependencies.txt
```diff
@@ -9,18 +10,21 @@ unixbench_order = 50
...
 unixbench-ver = ANY
+git-ver = ANY
 ### END - Dependency versions ###
 
-### START - Dependency URLs ###
-### END - Dependency URLs ###
-
 ### START -  Dependency and method-specific command lines ###
+common-git-install-man = package_install git-all
+common-unixbench-install-man = git clone https://github.com/kdlucas/byte-unixbench.git
```
hinesmr: I don't mean to be pedantic here, but this should be a fully-functional set of copy-and-paste commands. What happens here is that CloudBench spits out this command to the user in the case of `man` (manual) mode. Thus, this needs to include the correct path (a home directory, perhaps) where the clone should go, as well as the complete set of commands to configure, compile and install the binary in a location which precisely matches where the benchmark script that you have above is expecting it to be.

chriskirkland: Fair point.  I'll make the install location explicit (code assumes `~`) and test the install instructions against `CentOS`, `Fedora`, and `Ubuntu`.

chriskirkland:
@hinesmr Made the install instructions much more explicit and tested in Docker against Fedora 24, Centos 6/7, and Ubuntu 14.04/16.04.  Please review at your earliest convenience.

hinesmr:
@chriskirkland Looks great. Thanks for your patience. As I mentioned, we're in a freeze for a minor release to SPEC Cloud v1.1, so @maugustosilva will merge this as soon as we get through that freeze.

chriskirkland:
@hinesmr No problem; thank you for your time.  How long do you think the code freeze will be in place?  Thanks

PR #132, opened 2016-09-19
SPEC Cloud v1.1: fixes for public clouds

hinesmr:
1. Tests on amazon stopped working because they suddenly restricted their userdata to be 16K and we were over that limit. This fixes that by removing a duplicate copy of the OpenVPN configuration file that was being loaded in data_ops.py. Userdata files should be following cloud-config syntax rather than being inlined into shell, which is handled in lib/clouds/shared_functions.py

2. Non-SPEC cloud images from IBM causes non-IBM users to fail to startup on Amazon when users have no access to IBM-owned images. This commit eliminates non-SPEC images from the template. If you are using Amazon, you need to use your own configuration file to override these images and not depend on the template. Only DigitalOcean, as of today, has company-supported public images available to the general public, but not Amazon.

3. VPN support was broken somewhere with a missing '-i' switch to the `sed` command.

4. GCE bug found during configure time.


salmanbaset:
tested on OpenStack using spec cloud. working successfully. merging.

PR #133, opened 2016-09-20
one-line fix to restore libcloud multi-tenancy

hinesmr:
I tried my best to get the super-class perfect, but I just missed a single line of code. =) Silly me.

But this is a good example of fixing a bug that's common to all libcloud adapters without having to modify everybody's adapter at the same time.


salmanbaset:
LGTM

PR #135, opened 2017-02-06
CB "walkthrough mode", and more

maugustosilva:
This is a large but painfully (and some might say, expensively :-) well tested pull request (all regression tests described in item 5 were executed in one way or another).

The main motivator for the changes here are the improvement of CB's usability and
"debuggability" for first time users.

1.) CORE: Connecting to a new "empty" (i.e., without any configured images) will now cause the CLI
to go into "walkthrough" mode, guiding the user into a multi-step process where he will
verify that new instances can be attached, verify that the orchestrator can actually reach
into the instances, configure and capture a "tinyvm" image to be used by "nullworkload" and
then deploy its first AI.

1.1) To this end, the CLI/API command/call "vmattach" was changed in the following way:
a) passing the parameter "check:\<image ID or name\>" as the "vm role" will cause CB to just
issue the cloud-specific call for instance creation (without even attempting to establish
network connectivity with the instance). Useful to debug problems with instance creation.
b) passing the "check:\<image ID or name\>:\<login name\>", will cause CB to create the instance
and then attempt to establish an SSH connection to it. Useful to debug instance network
connectivity AND pubkey image injection.
c) passing the "check:\<image ID or name\>:\<login name\>:\<ai type\>", will cause CB to create
the instance, attempt to establish an SSH connection, and then transfer a copy of the
CB code to it. At this point, an user can just login on the instance (using the helper
script cbssh (vm number)) and run the automated installer (./install -r workload --wks <ai type>)

1.2) Additionally the CLI/API command/call "vmcapture" was also changed to have a "image name"
parameter, allowing an user, after running the automated installer to capture it with the
appropriate cloud-specific image name described the VM's role.

1.3) Cloud-specific image identifiers can be either image names or images unique ids.

1.3.1) Images names can now have a global prefix applied to it. By setting a value for the
attribute "IMAGE_PREFIX" - part of the GLOBAL object VM_DEFAULTS - multiple users can have
their own "private" versions of the same image (e.g., (user1)_cb_ycsb and (user2)_cb_ycsb>)
on the same public cloud.

The aforementioned behavior was implemented for the following clouds: SoftLayer (slr),
Amazon EC2 (ec2), OpenStack (osk), "Pure" Docker/Swarm (pdm), "Pure" LXD (lxd),
Google Compute Engine (gce), Digital Ocean (do) and Kubernetes (kub).
To do: VMWare VCloud (vcd), "pure" Libvirt (plm and ftc, and CloudStack (csk)

2.) CORE: Changed the CLI/API command/call "aiattach", allowing the passing of a composite parameter
in place of the usual AI type. Passing "build:<image ID or name>:<AI type>" will instruct
CB to create all instances for a new AI using the specified image identifier (and thus
overriding whatever VM role -> cloud-specific image identifier mapping already configured),
then run the automated installer script for each instances - installing all required
dependencies on the fly - followed by the regular AI setup scripts. For instance,
"aiattach build:baseubuntu:iperf" will deploy two "baseubuntu" empty images, run
"~/cbtool/install --role workload --wks iperf", and only then proceed with the regular iperf
setup process.
This allows the deployment of a complete workload using only "blank" images. While not
recommended for large-scale systematic experimentation, this feature opens the possibility
 of a quicker and simpler method to try out workloads.
Please note that, at this time, not all workloads can be automatically built on the fly
(e.g., ibm_daytrader requires the - AFAIK - impossible to fully automate DB2 installation,
and coremark requires you to go through an authentication process before downloading the
actual code).

3.) ADAPTERS: Major code re-organization, covering all cloud adapters capable of supporting the new
"walkthrough" mode.

3.1) They now have all the same method names in the same sequence (some of the methods are,
evidently, just stubs for certain clouds, such as "discover_hosts" for ec2) :
get_description
connect
test_vmc_connection
check_networks
check_images
check_ssh
check_security_group
discover_hosts
vmccleanup
vmcregister
vmcunregister
vmcount
get_ip_address
get_instances
get_images
get_networks
is_vm_running
is_vm_ready
vm_placement
vvcreate
vvdestroy
vmcreate
vmdestroy
vmcapture
vmrunstate
vmmigrate
vmresize
imgdelete
aidefine
aiundefine

3.2) They now make heavy use of common functions/methods stored on lib/clouid/shared_functions.py,
saving literally hundreds of lines of code, while making the creation of new adapters faster
and easier.

3.3) Two new Cloud Adapters: LXD and Kubernetes

4.) VAPP: Multiple improvements and bugfixes for the multiple AI/VAPP types supported by
CB

4.1) A better, more generic method to handle attached volumes. Through the use of the common
shell function \"automount_data_dirs\" and the VApp-specific parameters  <ROLE>_DATA_DIR.
<ROLE>_DATA_FSTYP (just for attached volumes), <ROLE>_DATA_SIZE (just for ramdisk/tmpfs) and
<ROLE>_DATA_FILESERVER_IP/<ROLE>_DATA_FILESERVER_PATH (for nfs mounts), data volumes are
automatically mounted in the appropriate directory. This is done very early in the AI
deployment process (while the generic cb_post_boot.sh script is executed), meaning that they
will be fully available as mountpoints during the execution of the AI-specific setup scripts.
Tested with: nullworkload, fio, filebench, btest, ibm_daytrader, cassandra_ycsb, redis_ycsb, mongo_ycsb

4.2) Allow JAVA_HOME and HADOOP_HOME to be set as Application Instance parameters for YCSB
(Cassandra, Redis and MongoDB) and Hadoop (Giraph too). In case the specified path is not
 present, a reasonable attempt will be made to locate it on the instance's disk.

4.3) Added support for an automatically configured "Load Balancer" instance using HAProxy.
This cloud-specific image name is indicated on the "LB" VM role.

4.4) Created "Dockerfiles" for the following workloads: nullworkload,btest,fio,ddgen,
filebench,iperf,netperf,nuttcp,xping,unixbench,cassandra_ycsb,redis_ycsb,mongo_ycsb,
hadoop,giraph,hpcc,linpack,parboil. Some workloads, such as coremark, linpack and parboil
will require that certain specific packages are pre-downloaded and made available on
cbtool/docker/workload, once this is done the helper script cbtool/docker/build_all.sh can
be invoked with the mandatory parameter <DOCKER REPO TAG> build all aforementioned workloads
as Dockers. Building with additional parameters "quiet push" will automatically push all the
images to the specified <DOCKER REPO TAG>. Please note that, for the time being, only Ubuntu-based
images are covered (I do not have the bandwidth to maintain both Ubuntu and RHEL/CentOS/Fedora
images, but it should be trivial for others to do so and contribute back).

4.5) A new AI-type, "parboil"

5.) CODE QUALITY: last but in no way shape or form least, more regression tests

5.1) The "core code" regression test, which is normally executed against a simulated cloud
and takes around 30 minutes to run:
a) cd cbtool/regression; ./regression_test.py make;
b) cbtool/cb --soft_reset --config configs/regression/simulated.txt --trace regression/regression_test_experiment_plan.txt 2>&1 > ~/teste.txt
c) ./regression_test.py validate ~/teste.txt

5.2) The "real multi-cloud" regression test, which tests the functionality of the cloud adapter
by:
I) connecting to the cloud
II) creating a single instance (without checking network connectivity to the instance)
III) creating a second instance and checking if the ssh pubkey was properly injected
IV) creating a third instance with a volume attached
V) capturing this last instance with the image name "regressiontest"
VI) creating a fourth instance using the newly captured "regressiontest" image name
VII) creating a fifth instance pointing to a non-existent image (should fail)
VIII) creating a sixt instance with a valid image name, but forcing a failure at the very end
of the creation.

After each test, CB checks if the number of images reported on its Object Store (Redis) and
directly by querying the cloud is the same, and then removes the instance in case of success.
The last two tests ensure that instance cleanup upon failure is being properly executed.

In order to run this regression test, perform the following steps :
a) mkdir ~/multicloud_regression; cp cbtool/configs/regression/* ~/multicloud_regression/
b) modify the files there with your credentials and IPs.
d) cd cbtool/regression; ./real_multicloud_regression.py configs/../../multicloud_regression/ sim,pcm,pdm,nop,osk,ec2,gce,do,slr high
e) just inspect the output in tabular format, looking for "PASS" for every line and column

The full testing of all clouds takes around 1.5 hours.

5.3) The "real application" regression test, which will, given an already attached cloud,
attempt to deploy each AI type listed, collect the management metrics samples, collect at
least 3 application (runtime) metrics, then remove the AI (and collect the detachment management
metric samples).

In order to run this regression test, perform the following steps:
a) cd cbtoo/regression
b) ./real_application_regression_test.py nullworkload,btest,fio,ddgen,filebench,iperf,netperf,nuttcp,xping,unixbench,cassandra_yscb,redis_ycsb,mongodb_ycsb,hadoop,giraph,hpcc,linpack,parboil
c) just inspect the output in tabular format, looking for "PASS" for every line and column

The full testing of all workloads takes around 1.2 hours (when using the relatively fast
"pure" Docker adapter). #


hinesmr:
@maugustosilva 

Wow, you have been busy. Did you do all this during christmas-time? lol.

First-pass questions:
1. For the global image prefix feature: Does "user" == "tenant"? Is this allowing one to use CB multi-tenancy without needing to share images across different tenant accounts? Was that the intended use of it?
2.

lib/clouds/libcloud_common.py
```diff
@@ -46,162 +46,108 @@ class LibcloudCmds(CommonCloudFunctions) :
...
             NOTE: FOO_SSH_KEY_NAME != FOO_SSH_KEY_NAME. The first one is the private key used by cbtool,
                   and the second one is the public key.
-
             NOTE: If your cloud doesn't support cloud-init, then you will need to make sure the VM image is prepared in advance
                   with the public key that corresponds to FOO_SSH_KEY_NAME
-
     @use_cloud_init: (Optional, Default False)
             NOTE: If your cloud doesn't support cloud-init, you will need to make sure your images have baked in the SSH public key on your own.
-
-    @use_volumes: (Optional, Default False)
```
hinesmr: Where did this go?

lib/clouds/libcloud_common.py
```diff
@@ -46,162 +46,108 @@ class LibcloudCmds(CommonCloudFunctions) :
...
     @tldomain: (Optional, Default False)
             The FQDN of your cloud. None if false.
-
     @extra: (Optional, Default empty)
             Extra fixed parameters to be used by libcloud that we don't know about, which get passed at instance creation time.
-
     '''
 
     @trace
-    def __init__ (self, pid, osci, expid = None, description = "OverrideMe", num_credentials = 2, use_ssh_keys = False, use_cloud_init = False, use_volumes = False, use_locations = True, use_sizes = True, tldomain = False, verify_ssl = True, extra = {}) :
```
hinesmr: Volumes removed again?

lib/clouds/libcloud_common.py
```diff
@@ -46,162 +46,108 @@ class LibcloudCmds(CommonCloudFunctions) :
...
         self.ft_supported = False
         self.current_token = 0
         self.cache_mutex = threading.Lock()
         self.token_mutex = threading.Lock()
-        self.description = description
+        self.provider = provider
         self.num_credentials = num_credentials
         self.use_ssh_keys = use_ssh_keys
         self.use_cloud_init = use_cloud_init
-        self.use_volumes = use_volumes
```
hinesmr: ditto

lib/clouds/libcloud_common.py
```diff
@@ -571,166 +610,329 @@ def get_vm_instance(self, obj_attr_list) :
...
+
+                obj_attr_list["cloud_vv_name"] = obj_attr_list["cloud_vv_name"].lower()
+                
+                obj_attr_list["last_known_state"] = "about to send volume create request"
+
+                self.common_messages("VV", obj_attr_list, "creating", _status, _fmsg)
+                
+                _mark1 = int(time())
+
+                _volume = connection.create_volume(int(obj_attr_list["cloud_vv"]),
```
hinesmr: Alright, so the volume conditionals need to be brought back. The purpose of those conditionals was for clouds like VCloud whose clouds don't support volumes at all within the libcloud driver for each particular cloud. This needs to be hard-coded and has no bearing on CLOUD_VV=X.

lib/clouds/libcloud_common.py
```diff
@@ -571,166 +610,329 @@ def get_vm_instance(self, obj_attr_list) :
...
-                location = location,
+                size = _size, 
+                location = _location,
                 ex_user_data = self.populate_cloudconfig(obj_attr_list) if self.use_cloud_init else False,
                 ex_create_attr = extra,
                 **kwargs
-            )
-
-            _time_mark_prs = int(time())
-            obj_attr_list["mgt_002_provisioning_request_sent"] = _time_mark_prs - int(obj_attr_list["mgt_001_provisioning_request_originated"])
```
hinesmr: Where did this go? Has it been unified somewhere?

lib/clouds/libcloud_common.py
```diff
@@ -820,143 +975,125 @@ def vmcreate(self, obj_attr_list) :
...
-            _msg += " (cloud-assigned uuid " + obj_attr_list["cloud_vm_uuid"] + ")"
-            _msg += "...."
-            cbdebug(_msg, True)
+            cbdebug("Last known state: " + str(obj_attr_list["last_known_state"]))
 
-            firsttime = True
-            _time_mark_drs = int(time())
-            while True :
-                _errmsg = "get_vm_instance"
-                cbdebug("Getting instance...")
```
hinesmr: This logic was very sensitive and difficult to get right in testing due to the way the libcloud adapter works. Was this logic preserved somewhere?

lib/clouds/libcloud_common.py
```diff
@@ -820,143 +975,125 @@ def vmcreate(self, obj_attr_list) :
...
+                obj_attr_list["mgt_902_deprovisioning_request_sent"] = int(time()) - int(obj_attr_list["mgt_901_deprovisioning_request_originated"])
+            
+            else :
+    
+                self.common_messages("VM", obj_attr_list, "destroying", 0, '')
+    
+                firsttime = True
+                _time_mark_drs = int(time())
+                _instance = self.get_instances(obj_attr_list)
+                while _instance and _curr_tries < _max_tries :
```
hinesmr: Oh, I see, nevermind. You just moved it down.

lib/clouds/libcloud_common.py
```diff
@@ -1103,48 +1221,81 @@ def vmrunstate(self, obj_attr_list) :
...
+            obj_attr_list["credentials_list"] = _credentials_list
+
+            _image_instance = self.get_images(obj_attr_list)
+                
+            if _image_instance :
+
+                _x = obj_attr_list["imageid1"] 
+                obj_attr_list["imageid1"] = _image_instance.id
+                
+                LibcloudCmds.catalogs.cbtool[_credentials_list].delete_image(_image_instance)
```
hinesmr: This actually works?

lib/clouds/shared_functions.py
```diff
@@ -813,13 +822,342 @@ def populate_interface(self, obj_attr_list) :
...
+
+        if self.get_description() == "SoftLayer Cloud" :
+            if len(imageid) == 7 and is_number(imageid) :
+                return True
+
+        if self.get_description() == "Google Compute Engine" :
+            if len(imageid) == 19 and is_number(imageid) :
+                return True
+
+        if self.get_description() == "DigitalOcean Cloud" :
```
hinesmr: This is a pretty crazy set of conditionals. Can't this be in a configuration file instead of being hard-code like this? If we implement a new cloud adapter it means I would have to go into this file and manually perform steps here that are not contained inside the individual cloud adapter.

lib/clouds/shared_functions.py
```diff
@@ -813,13 +822,342 @@ def populate_interface(self, obj_attr_list) :
...
+
+            if not _key_contents :
+                _fmsg = _key_type 
+                cberr(_fmsg, True)
+                return False
+            
+            _key_pair_found = False
+
+            _registered_key_pairs = {}
+            if self.get_description() == "Cloudbench SimCloud" or \
```
hinesmr: Same thing here. We can't have each individual cloud's name hard-coded in here like this. These need to into the individual adapters so that the maintainer of each adapter only needs to be responsible for what's in their head.

lib/clouds/shared_functions.py
```diff
@@ -813,13 +822,342 @@ def populate_interface(self, obj_attr_list) :
...
+
+    @trace
+    def check_security_group(self,vmc_name, security_group_name) :
+        '''
+        TBD
+        '''
+    
+        _security_group_name = False
+        
+        if security_group_name :
```
hinesmr: .... again.

lib/clouds/vcd_cloud_ops.py
```diff
@@ -17,73 +17,837 @@
...
-class VcdCmds(LibcloudCmds) :
+import libcloud.security
+
+class VcdCmds(CommonCloudFunctions) :
+    '''
+    TBD
+    '''
     @trace
     def __init__ (self, pid, osci, expid = None) :
-        LibcloudCmds.__init__(self, pid, osci, expid = expid, \
```
hinesmr: The behavior of all these parameters was intentional, to separate default libcloud functionality from configuration files. Please bring this back. By having these parameters, users can implement very bare-bones libcloud implementations very quickly without having to populate a full configuration file and then bring options into a configuration file at a later date.

configs/templates/_digitalocean.txt
```diff
@@ -32,6 +32,10 @@ CREDENTIALS = $DO_CREDENTIALS
 SSH_KEY_NAME = $DO_SSH_KEY_NAME
 KEY_NAME = $DO_KEY_NAME
 LOGIN = $DO_LOGIN
+NETNAME = private
+CAPTURE_SUPPORTED = $True
```
ibmcb: so, for Digital Ocean, yes, I have implemented (and tested), the capture function.

configs/templates/_digitalocean.txt
```diff
@@ -43,10 +47,15 @@ SIZE = from_vm_template
 LEAVE_INSTANCE_ON_FAILURE = $False
 FORCE_FAILURE = $False
 DISABLE_TIMESYNC = $False
+IMAGE_PREFIX = $EMPTY
+IMAGE_SUFFIX = $EMPTY
 
 # We're not as big as amazon yet. Go easy on us please.
 [AI_DEFAULTS : DO_CLOUDCONFIG ]
 ATTACH_PARALLELISM = 1
+CAPTURE_SUPPORTED = $True
```
ibmcb: same as above...  I have implemented (and tested), the capture function.

lib/clouds/libcloud_common.py
ibmcb: for every adapter (including the upper-layer libcloud one), there is now a "create volume" function, which can be either "real" or "fake"... if the user specifies cloud_vv=<something> in a cloud which does not support volumes, then the "fake" version of the function will be called, which will not actually create a real volume, but it will not cause any error...maybe there is another problem that I am missing...

lib/clouds/libcloud_common.py
ibmcb: No, volumes creation is there, I have even tested that on Digital Ocean....

lib/clouds/libcloud_common.py
ibmcb: So, I removed the "use_volumes", due to the existence of the "fake" volume functions, but of course its entirely open for discussion...

lib/clouds/libcloud_common.py
ibmcb: I see... that is the case that you had in mind: a "libcloud-based" Adapter without support for volumes. Trying to understand the case a bit better. In such a cloud, if the user simply do not specify CLOUD_VV=<something>, he will not see any problem, right? The problem will arise if he, accidentally, actually specify it. Is my understanding correct?

lib/clouds/libcloud_common.py
ibmcb: It was just moved a few lines down.

lib/clouds/libcloud_common.py
ibmcb: yes, sure... I noticed the problem during the tests, yeah :-) I initially have removed it, and the put it back :-)

lib/clouds/libcloud_common.py
ibmcb: right

lib/clouds/libcloud_common.py
ibmcb: So, I tested the imagedelete with this code, and it worked... I could see the newly created image   properly disappear from the Digital Ocean dashboard...

lib/clouds/shared_functions.py
ibmcb: Hahahaha... you caught me! I realized, in the middle of the code re-organization, that it was far from the best way to implement these conditionals. If you are willing to live with that for a short while, I will make the cleanup of it a priority for the *next* PR. If you insist in having this corrected on *this* PR...well, I can't, in good conscience blame you :-)

lib/clouds/shared_functions.py
ibmcb: same as above! I cannot disagree :-)

lib/clouds/shared_functions.py
ibmcb: That is bad form, but in my evaluation is less egregious than the previous two. Would you be willing to let this one slide for now?

lib/clouds/vcd_cloud_ops.py
ibmcb: My apologies. I must have done something wrong with the VCloud adapter. Will just restore it at once!

ibmcb:
Just to summarize. Would you accept the PR with the restoration of the `@use_volumes: (Optional, Default False)` and the rectification of the mixup with the "lib/clouds/vcd_cloud_ops.py"? 
Full disclosure: a "yes" means that the (admittedly) "crazy set of conditionals" would be fixed later.

hinesmr:
I clicked approve on those conditions you mentioned above. They can be implemented later, but let's get them in before SPEC is ready for the next minor point release. I've completed the necessary Kmeans changes for Elasticity and will share them tomorrow --- so we should be planning for a new minor release ASAP.

ibmcb:
Thank you very much. I am merging now....

PR #138, opened 2017-04-18
Additionalworkloads

maugustosilva:
Please find attached the results of regression tests type 2 ("real multi-cloud") and type 3 ("real application")

[real_multicloud_regression_test.txt](https://github.com/ibmcb/cbtool/files/930270/real_multicloud_regression_test.txt)

[real_application_regression_test.txt](https://github.com/ibmcb/cbtool/files/930269/real_application_regression_test.txt)


configs/templates/_gce.txt
```diff
@@ -28,7 +28,7 @@ SSH_KEY_NAME = $GCE_SSH_KEY_NAME
 
 [MON_DEFAULTS : GCE_CLOUDCONFIG ]
 COLLECT_FROM_HOST = $False # impossible for amazon
-COLLECT_FROM_GUEST = $True
+COLLECT_FROM_GUEST = $False
```
hinesmr: Why make all of these default to False?

lib/clouds/libcloud_common.py
```diff
@@ -370,30 +373,32 @@ def vmccleanup(self, obj_attr_list) :
...
+                            for _volume in _volumes :
+                                if _volume.name.count("cb-" + obj_attr_list["username"]) :
+                                    try :
+                                        cbdebug("Destroying: " + _volume.name + " (" + tenant + ")", True)
+                                        _volume.destroy()
+                                    except :
+                                        pass
+                                    _running_volumes = True
+                                else :
+                                    _msg = "Cleaning up DigitalOcean. Ignoring volume: " + _volume.name
```
hinesmr: Hmmm, that should not say DigitalOcean. Forgot to remove that when I wrote this. I'll add it to my next PR.

lib/clouds/shared_functions.py
```diff
@@ -869,6 +872,9 @@ def is_cloud_image_uuid(self, imageid) :
         if self.get_description() == "Parallel Container Manager Cloud" :
```
hinesmr: Weren't we going to get these hard-coded cloud sections out of here and into configuration files? Can we do that before merging this?

lib/clouds/vcd_cloud_ops.py
```diff
@@ -12,842 +13,85 @@
...
 from lib.auxiliary.code_instrumentation import trace, cbdebug, cberr, cbwarn, cbinfo, cbcrit
-from lib.auxiliary.data_ops import str2dic, DataOpsException
-from lib.remote.network_functions import Nethashget
-
-from shared_functions import CldOpsException, CommonCloudFunctions 
-
-from libcloud.compute.types import Provider
-from libcloud.compute.providers import get_driver
-from libcloud.compute.types import NodeState
+from libcloud_common import LibcloudCmds
```
hinesmr: This was merged a long time ago..... did it disappear somehow?

configs/templates/_gce.txt
maugustosilva: I was trying to be a minimalist here, and just follow the procedure adopted on all other adapters (please note that this - COLLECT_FROM_GUEST = $False - is the case for all cloud adapters). Given that deployments might fail due to errors on the restarting of gmond, leave this option disabled by default. The goal here is to have an AI up and running as fast as possible, without any errors, and only then activate more "features" (such as Guest OS resource metrics) gradually.

lib/clouds/shared_functions.py
maugustosilva: You're absolutely right. I have promised to do a re-factor on the code to remove all these. I confess that I did not finished it yet, due to the fact that will be more difficult than I initially have thought (it will involve the creation of cloud-specific ssh and security group management methods). If you agree, I would prefer to merge this "workload-centric" PR before performing an "cloud adapter-centric" refactor.

lib/clouds/vcd_cloud_ops.py
maugustosilva: My apologies, it was clearly my fault, when attempting to reconcile a conflict from a previous merge.

configs/templates/_gce.txt
hinesmr: Fair enough. =)

lib/clouds/shared_functions.py
hinesmr: ok, I'll hold you to that. 😃

PR #150, opened 2017-10-04
Large PR. Lots of enhancements and bug fixes

hinesmr:
- Feature: We now have round-robin scheduling of AI objects (atomically) between VMC objects. This feature allows you to spread AIs across multiple regions if multiple regions are attached at the same time. It can be host-specific if regions only contain one host in the case of private clouds, but the implementation is host-agnostic. Note: This feature does not spread VMs within an AI, but instead all of the VMs of a complete AI get assigned together. This is similar to the way multi-tenancy works, but assigning an entire AI to one VMC or another instead of just one tenant or another.
- SPEC Cloud: Hadoop now exports the number of map reduces / iterations it took to complete a kmeans clustering run.
- Enhancement: Guest OS data collection now includes all CPU metrics, including steal time (instead of just user and system time).
- Enhancement: GCE now supports OpenVPN in addition to Amazon and DigitalOcean.
- Enhancement: You can now add a semi-colon separated list of packages to install at the end of a role in the [VM_TEMPLATES] section which will automatically be handled by cloud-init. (See the DigitalOcean templates for examples)
- Enhancement: The libcloud adapter now supports all vmrunstate commands (stop, start, etc....).
- Enhancement: Rsyslog templating now supports better search/replace functions for more complex templates (such as relaying logs to remote syslog servers in addition to just local logging)
- Enhancement: Move the CB history file to the home directory so that it persists updates to the codebase.
- Fix / SPEC Cloud: GUI was not using the Experiement ID properly to separate data reported by multiple, simultaneous CloudBench installations communicating with the same MongoDB cluster.
- Fix: We have been dropping the metric store unconditionally even when using soft reset for a very long time unknowingly. This was evident when I started having multiple benchmarks point to the same cluster. That's now finally fixed.
- Fix: Logging format of both scripts and python needs to confirm to RFC-3164 (basically including the hostname in the wire protocol)
- Fix: Public DigitalOcean images upgraded to 16.04 LTS
- Fix: SSH-key discovery has an opt-out option to allow the discovery to not abort if the user has already pre-defined keys.
- Fix: SSH-key discovery no longer deletes keys. (Can be redesigned later)
- Fix: GUI was not handling floating point values correctly in some cases.
- Fix: GCE was missing a 'scope' option during initial use of your GCE credentials.
- Fix: The libcloud adapter was broken in a few ways. All fixed.
- Fix: CLOUD_VV should accept "False" as an option.
- Fix: The AIDRS 'arrived' counter was not counting properly.
- Fix: 'sudo su' was removed from process_operations.py for local commands.
- Fix: Retire background operations. Not used anywhere anymore.
- Little bit of return code cleanup in Redis.


hinesmr:
@maugustosilva We have a green light on all of these patches against SPEC Cloud using all three public cloud providers .... benchmark is functional. Just waiting for your review.

maugustosilva:
Done.  Thank you Michael.

PR #153, opened 2017-10-11
Pikefixes

maugustosilva:
Bugfix: OpenStack adapter now can connect to Ocata/Pike releases using
the latest python clients (e.g., python-novaclient==9.1.1). This includes
the ability to manage Floating IPs and create Jumphosts

Bugfix: Unixbench workload were not properly reporting application performance metrics

Bugfix: Make sure that entry that points at the VM’s hostname is
available on /etc/hosts during image creation (scripts `pre_install.sh` and `cb_cleanup`

Bugfix: Small fixes on the real multi-cloud regression


hinesmr:
:shipit:  ship it

PR #155, opened 2017-10-24
Gobindings and logging changes

hinesmr:
1. Introduce golang bindings for CloudBench. (yay)

2. This commit also includes a slight modification to
   the way logging works in "scalable" mode. Until now,
   when using scalable mode, logs from the VMs were disabled.
   But, Marcio recently introduced "~/cb_os_cache.txt" some
   time ago (very nifty) which saves us the network latency and
   no longer requires querying Redis on every log entry. This commit
   uses that functionality and restores logging in scalable mode.

### DIFF
--- a/clients/list_vms.go
+++ b/clients/list_vms.go
@@ -1,0 +1,52 @@
+/*
+# Copyright (c) 2017 DigitalOcean, Inc. 
+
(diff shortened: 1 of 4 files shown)


hinesmr:
@maugustosilva ping :)

PR #162, opened 2018-01-23
Multiple features and bugfixes

maugustosilva:
1) CODE QUALITY: Removed any cloud-specific references on the code in
                 lib/cloud/shared_functions.py. Now, adapter-specific
                 SSH key management (and Security Groups listing) code
                 was moved to cloud-specific adapter files.

2) CODE QUALITY: Relevant changes to the current libcloud adapter,
                 making it more configurable for use with multiple
                 adapters. Adding new libcloud backends should be
                 significantly easier now.

3) CORE:         At long last, create a clean, integrated
                 implementation for cloud-init generation that can
                 be used by all adapters. Two new attributes for
                 VM_DEFAULTS are defined:
				     USERDATA_POST_BOOT : if set to
				     $True, will cause the
				     cb_post_boot.sh script (e.g.,
				     containing an rsync pull from
				     the instance) to be executed
				     via cloud-init.
				     USERDATA_SSH : if set to
				     $True, will inject the
				     defined SSH key (specified by
				     the instance's SSH_KEY_NAME
				     attribute) through
				     cloud-init.

4) CORE:         Implemented a more robust, cloud-agnostic instance
                 VPN management. This is fully integrated with the
                 above mentioned attributes - USERDATA_POST_BOOT
                 and USERDATA_SSH - and was tested on both OpenStack
                 and Digital Ocean. Just as a reminder, here are the
                 VM_DEFAULTS attributes :
                     USE_VPN_IP: if set to
                     $True, the cloud-init script
                     will FIRST execute the OpenVPN
                     client to connect to the VPN
                     server, and then execute the
                     regular instance post boot
                     steps. However all VM initiated
                     traffic to the Stores (in the
                     Orchestrator) is done through
                     "normal", "outside of the cloud"
                     IP addresses.

5) ADAPTERS: New cloud adapter: Azure Service Management (ASM),
             using libcloud.

6) WORKLOADS: New workload, "mongo_acmeair" (AirLine travel
              reservation application).

7) BUGFIX: Fixed a bug that prevented the rsync normally done
           during instance post-boot to complete in case of
           an image with incorrect permission. This was solved by
           adding the option "--rsync-path='sudo rsync'" to the rsync.


lib/clouds/libcloud_common.py
```diff
@@ -542,13 +630,17 @@ def get_ip_address(self, obj_attr_list) :
             _status = 100
             node = self.get_instances(obj_attr_list)
 
-            if len(node.private_ips) > 0 and obj_attr_list["run_netname"].lower() == "private" :
-                obj_attr_list["run_cloud_ip"] = node.private_ips[0]
+            if len(node.private_ips) > 0 :
```
hinesmr: As we were discussing on mattermost, we really have to stop the key/value bleeding. I'm totally fine with REMOVING all of these `use_xxxx` statements that I introduced, so long as we find a way to do this in a configuration file templates that is shared amongst all the libcloud-based adapters without requiring the user to have to define them all or even know about them all in advance to get started developing a new cloud adapter.

lib/clouds/libcloud_common.py
```diff
@@ -108,29 +122,37 @@ class LibcloudCmds(CommonCloudFunctions) :
...
-        self.use_cloud_init = use_cloud_init
         self.use_volumes = use_volumes
         self.use_locations = use_locations
+        self.use_services = use_services
+        self.use_networks = use_networks
+        self.use_security_groups = use_security_groups
+        self.use_floating_ips = use_floating_ips
+        self.use_public_ips = use_public_ips        
         self.use_sizes = use_sizes
+        self.use_get_image = use_get_image
```
hinesmr: As we were discussing on mattermost, we really have to stop the key/value bleeding. I'm totally fine with REMOVING all of these use_xxxx statements that I introduced, so long as we find a way to do this in a configuration file templates that is shared amongst all the libcloud-based adapters without requiring the user to have to define them all or even know about them all in advance to get started developing a new cloud adapter.

lib/clouds/libcloud_common.py
```diff
@@ -886,89 +981,114 @@ def vmcreate(self, obj_attr_list) :
...
+            _location = self.get_region_from_vmc_name(obj_attr_list)
+            
             extra.update(self.pre_vmcreate(obj_attr_list, extra))
 
-            if self.use_locations :
-                _location = [x for x in LibcloudCmds.locations if x.id == obj_attr_list["region"]][0]
-            else :
-                _location = False 
-
+            obj_attr_list["libcloud_size_inst"] = False
```
hinesmr: Please make use of pre_vmcreate. This is were modifications to the dictionary should happen, via callsbacks, as opposed to additionally stubbed functions. If the callback doesn't exist, then it just becomes a no-op in the cloud's adapter without needing to define the stub at all.

lib/clouds/shared_functions.py
```diff
@@ -230,22 +230,22 @@ def wait_for_instance_ready(self, obj_attr_list, time_mark_prs) :
...
                 _msg += " s) is actually longer than the "
                 _msg += "interval between pooling attempts (" + str(_wait) + " s)."
                 cbdebug(_msg, True)
                 _actual_wait = 0
 
-            # There is still some reconciliation to be done here. If vpn_only is used, then only openvpn-initiated callbacks should set the pending attribute, not userdata scripts. There is a distinction. It also means that public_cloud_ip should be set as well and that access to pending attributes is a requirement. See get_ip_address from do_cloud_ops.py
-            # Also "use_vpn_ip" is used throughout the scripts and codebase, so use_vpn_ip is already reserved to ensure that vpn_only works as it did before.
-            # So any changes to use_vpn_ip need to be conditionalized with an extra check to vpn_only as well.
-            if str(obj_attr_list["use_vpn_ip"]).lower() != "false" and str(obj_attr_list["vpn_only"]).lower() == "false" :
+            if str(obj_attr_list["use_vpn_ip"]).lower() != "false" :
```
hinesmr: Is "VPN_ONLY" deprecated? or implied now? How does that work?

lib/clouds/shared_functions.py
```diff
@@ -825,64 +825,9 @@ def populate_interface(self, obj_attr_list) :
                     except Exception, msg :
                         _fmsg = "Could not lookup interface " + iface + " for hostname " + hostname + " (probably bad /etc/hosts): " + str(msg)
                         raise CldOpsException(_fmsg, 1295)
-                    
-    @trace
-    def is_cloud_image_uuid(self, imageid) :
```
hinesmr: Why are these checks necessary in the first place? Why does it matter that image IDs can be uuids or not? Is this specific to OpenStack? And if so, could we just rip out the whole thing and keep it contained to openstack?

lib/clouds/shared_functions.py
```diff
@@ -1187,33 +1021,31 @@ def base_check_images(self, vmc_name, vm_templates, registered_imageid_list, map
...
-"""
-            for line in obj_attr_list["userdata"].split("\n")[:-1] :
-                cloudconfig += "      " + line + "\n"
-
-        # We need the VPN's IP address in advance, which was solved before
-        # the previous VPN support was gutted, but since we're left to do it
-        # on our own, we need cloud-config to send our VPN configuration file
-        # in advance.
-        conf_destination = "/etc/openvpn/" + obj_attr_list["cloud_name"] + "_client-cb-openvpn-cloud.conf"
+        cloudconfig = "#cloud-config\n"
```
hinesmr: Please bring back the multi-line string syntax. It makes understanding cloud-config much easier.

If you were not aware in python, multi-line strings can be appended to `on-the-fly` in python by simply concatenating regular strings to multi-line strings within the same syntax.

lib/clouds/shared_functions.py
```diff
@@ -1247,38 +1079,115 @@ def populate_cloudconfig(self, obj_attr_list) :
...
-  - chmod +x /etc/openvpn/client_connected.sh
-  - mv """ + conf_destination + """ /tmp/cbvpn.conf
-  - rm -f /etc/openvpn/*.conf /etc/openvpn/*.ovpn
-  - mv /tmp/cbvpn.conf """ + conf_destination + """
-  - mkdir -p /var/log/openvpn
-  - openvpn --daemon --config """ + conf_destination + """
-"""
-  #- systemctl start openvpn@""" + obj_attr_list["cloud_name"] + """_client-cb-openvpn-cloud.service
-  #- service openvpn start
+        if obj_attr_list["userdata_post_boot"].lower() == "true" or obj_attr_list["use_vpn_ip"].lower() != "false" :
```
hinesmr: Same thing here. Using explicit newlines is prone to error for both of us, and it becomes increasingly difficult to remember when to use a newline and when not to use a newline.

Multi-line strings have all the same abilities as regular strings.

lib/clouds/shared_functions.py
```diff
@@ -1247,38 +1079,115 @@ def populate_cloudconfig(self, obj_attr_list) :
...
+            cloudconfig += "\nruncmd:\n"
+            cloudconfig += "  - chmod +x /tmp/cb_post_boot.sh\n"
```
hinesmr: Why is this cb_post_boot? That is not the same as the real post_boot is it? Userdata size has a distinct limit on Amazon and making the userdata too large will cause the userdata upload to fail.

configs/regression/do.txt
```diff
@@ -8,7 +8,7 @@ STARTUP_CLOUD = TESTDO
 
 [USER-DEFINED : CLOUDOPTION_TESTDO ]
 DO_INITIAL_VMCS = nyc1:sut
-DO_CREDENTIALS = long_hex_token_from_digitalocean.com:arbitrarytag1,another_token:arbitrarytag2,another_token:arbitrarytag3
+DO_CREDENTIALS = arbitrarytag1:long_hex_token_from_digitalocean.com,arbitrarytag2:another_token
```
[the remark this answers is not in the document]
maugustosilva: Yes, will put it back

configs/templates/_do.txt
```diff
@@ -0,0 +1,102 @@
+[USER-DEFINED]
```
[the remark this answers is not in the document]
maugustosilva: Mistake.... there should be only one. The other will be removed

configs/templates/_ftc.txt
```diff
@@ -92,63 +92,65 @@ LOGIN = $FTC_LOGIN
 SSH_KEY_NAME = $FTC_SSH_KEY_NAME
 
 [VM_TEMPLATES : FTC_CLOUDCONFIG]
+ACMEAIR = size:iron32, eclipsed_size:gold32, imageids:3, imageid1:cloudbench_root, imageid2:cloudbench_swap, imageid3:config_drive
```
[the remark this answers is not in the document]
maugustosilva: Great! Will do so

lib/api/api_service_client.go
```diff
@@ -43,15 +43,11 @@ func (api *APIClient) Call(method string, args ...interface{}) (result map[strin
 		return nil, errors.New("Failed to provide CloudBench URL.")
     }
 
-    response, err := xmlrpc.Request(api.Address, method, args...)
+    response := xmlrpc.Request(api.Address, method, args...)
```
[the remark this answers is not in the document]
maugustosilva: My apologies

lib/clouds/as_cloud_ops.py
```diff
@@ -0,0 +1,151 @@
...
+            _cloud_service_found = True
+
+        hash_object = hashlib.sha1(str(time()))
+        hex_dig = hash_object.hexdigest()
+        vm_defaults["vm_name_suffix"] = str(hex_dig[0:10])
+
+        return _cloud_service_found
+
+    @trace
+    def get_list_node_args(self, obj_attr_list) :
```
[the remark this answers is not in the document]
maugustosilva: This `obj_attr_list["cloud_service_name"] ` is also Azure-specific. I advance here the same proposal as above (leave as is for now, and will fix it later).

lib/clouds/as_cloud_ops.py
```diff
@@ -0,0 +1,151 @@
...
+                return _service.service_name
+
+        return False
+    
+    @trace
+    def get_cloud_specific_parameters(self, obj_attr_list, extra, credentials_list, status) :
+        '''
+        TBD
+        '''
+        self.vmcreate_kwargs["ex_custom_data"] = obj_attr_list["userdata"]
```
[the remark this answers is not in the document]
maugustosilva: Normally, I would put it there, but this requires information that is only available afterwards (for instance, the result of `populate_cloudconfig`).  I certainly don't like  what I had to do here, but not clear how to solve it. Would you be willing to accept it as technical debt?

lib/clouds/asm_cloud_ops.py
```diff
@@ -0,0 +1,134 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#/*******************************************************************************
+
+'''
+    Created on Oct 31, 2015
```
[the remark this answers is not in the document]
maugustosilva: My mistake while renaming the adapter. Will be removed.

lib/clouds/asm_cloud_ops.py
```diff
@@ -0,0 +1,134 @@
...
+from time import time
+import hashlib
+
+#from libcloud.compute.drivers.azure import ConfigurationSet,ConfigurationSetInputEndpoint
+
+class AsmCmds(LibcloudCmds) :
+    @trace
+    def __init__ (self, pid, osci, expid = None) :
+        LibcloudCmds.__init__(self, pid, osci, expid = expid, \
+                              provider = "AZURE", \
```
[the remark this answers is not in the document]
maugustosilva: My mistake while renaming the adapter. Will be removed.

lib/clouds/do_cloud_ops.py
```diff
@@ -15,29 +15,114 @@
...
+        '''
+        TBD
+        '''
+        
         driver = libcloud_driver(access_token, api_version = 'v2')
 
         return driver
+
+    @trace
+    def extra_vmc_setup(self, vmc_name, vmc_defaults, vm_defaults, vm_templates, connection) :
```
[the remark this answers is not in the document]
maugustosilva: Will do.

lib/clouds/do_cloud_ops.py
```diff
@@ -15,29 +15,114 @@
...
+    
+    @trace
+    def get_cloud_specific_parameters(self, obj_attr_list, extra, credentials_list, status) :
+        '''
+        TBD
+        '''
+        _mark_a = time()
+        keys = []
+
+        tmp_keys = obj_attr_list["key_name"].split(",")
```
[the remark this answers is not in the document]
maugustosilva: Make sense. Will modify accordingly.

lib/clouds/do_cloud_ops.py
```diff
@@ -15,29 +15,114 @@
...
+        self.annotate_time_breakdown(obj_attr_list, "get_sshkey_time", _mark_a)
+                
+        extra["ssh_keys"] = keys
+        
+        if obj_attr_list["netname"] == "private" :
+            extra["private_networking"] = True
+
+        obj_attr_list["libcloud_call_type"] = 1
+
+        self.vmcreate_kwargs["ex_create_attr"] = extra
```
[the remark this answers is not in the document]
maugustosilva: Please see my comment on `self.vmcreate_kwargs["ex_custom_data"] = obj_attr_list["userdata"]`.

lib/clouds/gce_cloud_ops.py
```diff
@@ -706,6 +744,29 @@ def is_vm_ready(self, obj_attr_list) :
             obj_attr_list["last_known_state"] = "not running"
             return False    
 
+    @trace
+    def vm_placement(self, obj_attr_list) :
```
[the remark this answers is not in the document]
maugustosilva: Will gradually remove stubs, for sure. During the major adapter re-design and cleanup that happened last year, the stubs were useful for me to organize the the adapters were all "isomorphic". With your agreement, will put this on "technical debt".

lib/clouds/libcloud_common.py
maugustosilva: I am ok on removing these too, over time.

lib/clouds/libcloud_common.py
maugustosilva: I don't think that I got that... "run_netname" and "prov_netname" are common keys for all clouds, and I depend on then heavily, specially on "non-VM" clouds, such as Kubernetes, Docker and LXD. Removing these would require, in my evaluation, a significant re-design of all adapters.

lib/clouds/libcloud_common.py
maugustosilva: Please see my comment on `self.vmcreate_kwargs["ex_custom_data"] = obj_attr_list["userdata"].` (i.e., `vmcreate_kwargs` requires information that is only available after `pre_vmcreate`).

lib/clouds/shared_functions.py
maugustosilva: No, not deprecated at all! VPN_ONLY is used when deciding the IP address of the Object Store and the File Store. Please look at the `create_bootstrap_script` method in "shared_function.py"

lib/clouds/shared_functions.py
maugustosilva: This is actually used by OpenStack, LXD and Docker, among others. In addition to it, I am using this function to "translate" the images names (on `VM_TEMPLATES : <MODEL>_CLOUDCONFIG]`) into image uuids. I would really like to keep it, unless you strongly object to it. I am willing to even go and put it on "technical debt" if you insist on removal.

lib/clouds/shared_functions.py
maugustosilva: yeah, ok... I actually like the single-line string syntax (this is just my personal preference), but you object and requires multi-line syntax, I will bring those back, sure.

lib/clouds/shared_functions.py
maugustosilva: Same comment as above. Will bring it back if you insist, no problem at all.

lib/clouds/shared_functions.py
maugustosilva: So, one of the possible options (on a per-cloud basis) is to run the actual `cb_post_boot.sh` via cloud-init. It does not increase the size of the userdata, since all the actual data is obtained via an rsync (also on the cloud-init).

lib/clouds/libcloud_common.py
hinesmr: OK, so there was some detection logic in get_ip_address that was removed for making sure that `run_cloud_ip` points to the VPN IP address instead of the other IP address. What happened to it?

lib/clouds/shared_functions.py
hinesmr: I'll need to test this first using a separate PR.

lib/clouds/as_cloud_ops.py
```diff
@@ -0,0 +1,151 @@
...
+        vmc_defaults["cloud_service_name"] = vm_defaults["cloud_service_name"]
+        
+        _cloud_service_name_list = _local_conn.ex_list_cloud_services()
+        
+        for _cloud_service in _cloud_service_name_list :
+            if _cloud_service.service_name == vm_defaults["cloud_service_name"] :
+                _cloud_service_found = True
+
+        if not _cloud_service_found :
+            _local_conn.ex_create_cloud_service(vm_defaults["cloud_service_name"], vmc_name)
```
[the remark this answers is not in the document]
maugustosilva: This "ex_create_cloud_service" libcloud method only exists for Azure. Would you be willing to accept it as is, for the time being? I propose that we leave it as is for now, with my promise to go over and cleanup the libcloud adapter later, in a more integrated manner.

lib/clouds/shared_functions.py
hinesmr: I think you misunderstood. I was simply asking why you renamed it to post_boot......... because that is confusing with the original post_boot. I definitely don't want to run the original post boot over cloud-init. Let's keep it as is. Do you have a strong reason to rename the userscript to "post_boot"? Can you bring back the original name so that it does not clash with our current post boot script?

lib/clouds/as_cloud_ops.py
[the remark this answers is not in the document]
hinesmr: There's no sense of urgency here. The libcloud changes can be worked on more in their own branch/PR. Since nobody is explicitly using this adapter right now, we can try to get it right the first time.

PR #170, opened 2018-02-14
Support for VPN connection errors

hinesmr:
So, I'm not all that proud of this patch, but I had a hard time thinking of a better solution.

While floating IPs and private IPs stay the same, VPN addresses can (and do) change a lot.

On the orchestrator-side, we already have solutions for this in place that allow the orchestrator
to update it's IP address on-the-fly in case your laptop restarts or the connection resets.

But, on the VM-side, if a VPN connection resets, you're SOL without this patch.

The most important part of this patch is not just the VM-side that updates Redis with the
new IP, but really is the SSH-side of this which *dynamically* contacts Redis to get the
most recently-known IP address before each attempt to retry an SSH command.

Patching the retry support was not so straight forward: Now, instead of embedding the IP
address into the SSH command, we lookup that IP on-the-fly each time the ProcessManagement
code needs to repeat the command.

So, it's a little messy, but I tested it quite a bit and it seems stable.

I'm happy to re-work the patch a little bit if it needs any additional cleanups.

### DIFF
--- a/lib/clouds/libcloud_common.py
+++ b/lib/clouds/libcloud_common.py
@@ -897,10 +897,9 @@ class LibcloudCmds(CommonCloudFunctions) :
 
             if self.use_ssh_keys :
 
-                keys = []
-    
                 tmp_keys = obj_attr_list["key_name"].split(",")
                 for dontcare in range(0, 2) :
+                    keys = []
                     for tmp_key in tmp_keys :
                         for key in LibcloudCmds.keys[_credentials_list] :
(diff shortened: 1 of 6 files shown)


maugustosilva:
You are right, this patch seems messy :-). However it does not seem to break the anything else (e.g., the automated dependency installer). I do worry about these extra calls to redis to resolve the IP from the UUID, but if you tested it at scale, and reported it as stable, I consider a valid evidence. 

Being that the case, I will abide by my personal philosophy of "add (tested) functionality NOW, and then USE it for a while before revisiting and cleaning up code" and will merge. Thanks!

PR #181, opened 2018-03-27
Replace outdated add_floating_ip() with update_floatingip()

rayx:
Novaclient's add_floating_ip() API was removed in Queens release [1].
That causes cbtool fails to associate floating IP with VM when using
latest OpenStack release. The change replaces it with neutronclient's
update_floatingip().

[1] https://review.openstack.org/#/c/532999/


ibmcb:
Thanks!

PR #184, opened 2018-04-05
Use glanceclient to access OpenStack image API

rayx:
When creating VMs in parallel, cbtool creates novaclient client object
in main thread and invokes methods on that object in multiple worker
threads. This approach could cause random failures because novaclient
is not thread-safe. A main issue is that novaclient provides methods
to access image and network APIs and those methods change client
object's service_type attribute, which is one of the components used
to generate request URL. So for example, if a worker thread calls
image API when there are other threads calling compute APIs, the latter
would fail. I observed the issue consistently when I ran SPECcloud's
yscb baseline script (I replaced its cbtool with the latest vesion)
on systems which had latest version of novaclient installed.

The change works around the issue by using glanceclient to call image
API. With this change the issue is gone on my systems.

### DIFF
--- a/lib/clouds/osk_cloud_ops.py
+++ b/lib/clouds/osk_cloud_ops.py
@@ -44,6 +44,7 @@ except:
 
 from novaclient import client as novac
 from novaclient import exceptions as novaexceptions
+from glanceclient import client as glancec
 
 from lib.auxiliary.code_instrumentation import trace, cbdebug, cberr, cbwarn, cbinfo, cbcrit
 from lib.auxiliary.data_ops import str2dic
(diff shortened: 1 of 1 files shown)


hinesmr:
@maugustosilva ping?

rayx:
FYI. I have resolved the merge conflict in my local repo, but haven't got a chance to test it (including the recent large commit). Will update the PR once I get it done.

rayx:
I find the code change in the recent commit (beaf287d25b34750f10f8e0f5fe1566f732425f0) has somehow resolved the concurrency issue that the PR intended to fix. Before that commit, cbtool shared a single novaclient object in all threads. As some threads changed that object's state (e.g, service_type attribute), request URLs were mal-formed and hence failures. With the code change in that commit, cbtool creates a novaclient object for each VM. So, if there aren't multiple threads calling novaclient's API on the same VM, it should work fine. 

That said, I still think this PR has its value because a) it works even if there are multiple threads calling OpenStack API on the same VM (I don't know if the scenario exists in cbtool or not), and b) it has very low risk and have no impact on anything. So I'll continue to test it and update the PR (perhaps late next week).

rayx:
Thanks!

PR #189, opened 2018-04-25
Libcloud-based adapter for OpenStack.

maugustosilva:
Keeping up-to-date with changes in the myriad of OpenStack Clients
has been a problem for us for a long time. This update introduces
an alternative adapter for OpenStack clouds, leveraging the work
done on libcloud (OpenStack backend). For now, I propose we let the
two adapters to co-exist, and as we harden and mature this code, we
can obsolete the other adapter.


hinesmr:
Amazing. The OpenStack libcloud adapter is so short too!

PR #192, opened 2018-04-25
Just re-arranged to role list on VM Templates (for all clouds) in Alphabetical order

maugustosilva:
In anticipation for some changes on "VM Template" that I will propose
- the thought occurred while writing an "How to add new Workloads"
Guide (still not done) - I am at least re-organizing the list of
roles that one has to change (in all adapters) in alphabetical order.


configs/templates/_do.txt
```diff
@@ -40,8 +40,6 @@ NETNAME = $DO_NETNAME
 CAPTURE_SUPPORTED = $True
 RUNSTATE_SUPPORTED = $True
 RESIZE_SUPPORTED = $False
-PROV_NETNAME = $DO_NETNAME
-RUN_NETNAME = $DO_NETNAME
```
hinesmr: Where did these go?

configs/templates/_do.txt
```diff
@@ -83,22 +81,78 @@ SECURITY_GROUPS = not_yet_applicable
...
+ACMEAIR = size:4gb, imageid1:cb_acmeair
+APACHE = size:4gb, imageid1:cb_wrk
+BONNIE = size:2gb, imageid1:cb_bonnie
+BTEST = size:2gb, imageid1:cb_btest
+CAFFE = size:4gb, imageid1:cb_caffe
 CASSANDRA = size:4gb, imageids:1, imageid1:cloudbench-ycsb-cassandra-on-1604.060117-1
-YCSB = size:4gb, imageids:1, imageid1:cloudbench-ycsb-cassandra-on-1604.060117-1
-SEED = size:4gb, imageids:1, imageid1:cloudbench-ycsb-cassandra-on-1604.060117-1
+CHECK = size:512mb, imageids:1, imageid1:to_replace
+CLIENT_IBM_DAYTRADER = size:4gb, imageid1:cb_daytrader
```
hinesmr: Except for the `check` image, can we revert this change? These are not public images. 

If the user runs the workload builder (and we should encourage them to do so), then we might want to provide an "include" file after the builder is done that shows them how to define the new workloads.

But, otherwise, I would prefer we did not have workload definitions with images that don't really exist in the public cloud.

configs/templates/_ec2.txt
```diff
@@ -89,65 +99,67 @@ LOGIN = $EC2_LOGIN
...
-DRIVER_TRADELITE = size:t2.medium, imageid1:cb_tradelite
-MYSQL = size:t2.medium, lb_size:t2.large, imageid1:cb_open_daytrader
-GERONIMO = size:t2.medium, imageid1:cb_open_daytrader
+CLIENT_WINDOWS = size:t2.medium, imageid1:cb_windows
+CN_HPC = size:t2.medium, imageid1:cb_hpcc
+COREMARK = size:t2.micro, imageid1:cb_coremark
 DB2 = size:t2.medium, lb_size:t2.large, imageid1:cb_daytrader
-WAS = size:t2.medium, imageid1:cb_daytrader
+DDGEN = size:t2.medium, imageid1:cb_hpcc
+DRIVER_COREMARK = size:t2.micro, imageid1:cb_coremark
```
hinesmr: Same concern here for Amazon. Basically we kind of "spam" the console during initial CB startup with errors that are not really errors.

If the user hasn't defined any images yet (or hasn't used the workload builder), then `CHECK` really is the only image that needs to be here. 

I've been wanting to remove these for a long time....

maugustosilva:
Following our quick chat on Mattermost, I am merging this PR. Thank you, Michael

PR #212, opened 2018-05-25
Add mem_anonpages metric

rayx:
The change adds Gmond module mem_stats.py to collect mem_anonpages
data, and modifies internal_options.txt to include this metric. It
also removes mem_shared from internal_options.txt because it's obsolete
and useless (See [1]. /proc/meminfo doesn't has a field for MemShared.
As a result, Gmond always returns 0 as mem_shared's value).

The mem_stats.py script and its config file are mostly copied from
ibmcb/monitor-core repository. I try to keep them unchanged as much
as possible, but have to make a few changes:

- Removed mem_total, mem_free, mem_buffers, and mem_cached from
  mem_stats.py, because otherwise they would override metrics of the
  same names from Gmond built-in mem module. The override occurs because
  the gmond config file in VM has python_module placed before all
  other modules (see scripts/common/cb_create_gmond_config_file.sh).
  I guess the author had a reason to do that, so I don't change it.

- Added an entry for mem_anonpages in mem_stats.conf.

Please note that metrics provided by mem_stats.py are measured in Byte,
instead of KB. That means mem_anonpages has different unit than mem_toal,
etc., from Gmond mem module. I considered to change it but didn't,
because the change applies only to mem_stats.py installed in VMs (I
suppose user would install mem_stats.py on host node by using package
manager). It would be very confusing if mem_stats.py in VMs and on host
nodes return data using different units.

[1] https://github.com/torvalds/linux/blob/master/Documentation/filesystems/proc.txt


rayx:
Below are my changes to mem_stats.py and mem_stats.conf:

1) diff of mem_stats.py

 $ diff -U 1 3rd_party/monitor-core/gmond/python_modules/memory/mem_stats.py util/python_ganglia_modules/mem_stats.py 
--- 3rd_party/monitor-core/gmond/python_modules/memory/mem_stats.py	2018-05-07 02:27:07.848646029 +0000
+++ util/python_ganglia_modules/mem_stats.py	2018-05-17 06:10:10.809038825 +0000
@@ -62,30 +62,2 @@
 
-    descriptors.append( create_desc(Desc_Skel, {
-                "name"       : "mem_total",
-                "orig_name"  : "MemTotal",
-                "units"      : "Bytes",
-                "description": "Total usable ram",
-                }))
-
-    descriptors.append(create_desc(Desc_Skel, {
-                "name"       : "mem_free",
-                "orig_name"  : "MemFree",
-                "units"      : "Bytes",
-                "description": "The amount of physical RAM left unused by the system. ",
-                }))
-
-    descriptors.append(create_desc(Desc_Skel, {
-                "name"       : "mem_buffers",
-                "orig_name"  : "Buffers",
-                "units"      : "Bytes",
-                "description": "Buffers used",
-                }))
-
-    descriptors.append(create_desc(Desc_Skel, {
-                "name"       : "mem_cached",
-                "orig_name"  : "Cached",
-                "units"      : "Bytes",
-                "description": "Cached Memory",
-                }))
-
     descriptors.append(create_desc(Desc_Skel, {
@@ -112,3 +84,3 @@
     descriptors.append(create_desc(Desc_Skel, {
-                "name"       : "mem_total",
+                "name"       : "mem_active_anon",
                 "orig_name"  : "Active(anon)",
@@ -367,2 +339,2 @@
         v = d['call_back'](d['name'])
-        print 'value for %s is %f' % (d['name'],  v)
\ No newline at end of file
+        print 'value for %s is %f' % (d['name'],  v)

2) diff of mem_stats.conf

 $ diff -U 1 3rd_party/monitor-core/gmond/python_modules/conf.d/mem_stats.pyconf util/ganglia_conf.d/mem_stats.conf 
--- 3rd_party/monitor-core/gmond/python_modules/conf.d/mem_stats.pyconf	2018-05-07 02:27:07.844646076 +0000
+++ util/ganglia_conf.d/mem_stats.conf	2018-05-16 08:50:20.214614544 +0000
@@ -32,3 +32,8 @@
   }
-
+ 
+  metric {
+    name = "mem_anonpages"
+    title = "AnonPages"
+    value_threshold = 1.0
+  }
 }

rayx:
I uploaded an image to show why I think mem_anonpages is useful:

https://github.com/rayx/rayx.github.io/blob/master/tmp/cbtool_host_and_vm_memory_usage.png

The image shows host and all VMs' memory usage in an experiment in which I started two AIs on a single-node cloud (one was YCSB, another was HADOOP). The upmost graph is host memory usage. The rest are each VM's memory usage. The X-axis is time. 

The area in red color indicates memory used for mem_anonpages. It's clear that every time a new AI was created, host's mem_anonpages increased. 

Another benefit to include mem_anonpages metric is that the sum of mem_free, mem_buffers, mem_cached, and mem_anonpages is semi-constant (note it's not mem_total, which includes kernel stack and heap memory also). That makes the graphs easy to read. Take the upmost one (host memory usage) as an example, if the graph contained mem_free, mem_buffers, and mem_caches only, it would be hard to interpret what the data meant.

hinesmr:
Great PR. Thanks, Ray.

rayx:
Glad it's useful. Thanks.

PR #214, opened 2018-05-25
Fix: the FIO workload needs time_based

jdesfossez:
The fio profile is configured to be time based, but without the
time_based option in the profile, it can exit before the expected
runtime (when it has completely read/written the target file).


hinesmr:
That might be the record for not only the shortest PR, but also the one with the fewest characters, lol.

PR #215, opened 2018-05-30
Fix directory name conflict in Hadoop workload

rayx:
In commit 3fa27ac53d5f0682e8d91c9f7e632996e1e29922 a few new directory
variables (HADOOPSLAVE_DATA_DIR, HADOOPMASTER_DATA_DIR, etc.) are added
in virtual_application.txt of hadoop workload. These directories are
created and used as mount point by automount_data_dirs() in cb_post_boot.sh.
The issue is that these variables have the same value as DFS_DATA_DIR,
which causes datanode service fails to start, because automount_data_dirs()
creates these directories with sudo, but hadoop datanode service runs as
cbuser. While it's possible to modify script to change the directory's
owner to cbuser, I don't think that's a good approach because, if I
understand it correctly, HADOOP[SLAVE|MASTER]_DATA_DIR and DFS_DATA_DIR
are for different kinds of data and should be different. So the change
fixes the issue by renaming DFS_DATA_DIR and DFS_NAME_DIR.


rayx:
The PR is obsolete by PR #216. So I'll close it.

PR #223, opened 2018-06-08
Fix a few bugs with gmetad startup code

rayx:
After enabling COLLECT_FROM_HOST, I observed the issue consistenly if I
exited cbtool without detaching VMC or cloud and restarted it with --soft_reset
option:

  status: Host OS performance monitor daemon (gmetad.py) started successfully.
          The process id is [] (using ports 8653 and 8653).

This occurs due to a few bugs with gmetad startup code:

- One is with ProcessManagement.start_daemon(). Note that get_pid_from_cmdline()
  returns an empty list when it fails to find matching process. In that case,
  start_daemon() returns a list containing one empty list. The value is not
  considered empty in Python and makes cbtool mistakenly thinks the process
  was started successfully although it wasn't.

- Another is with ProcessManagement.get_free_port(). It allocates the same
  port when being called multiple times.

The change fixes the above bugs. It also kills gmetad left in prior session
when cbtool re-attaches to VMC.

### DIFF
--- a/lib/operations/active_operations.py
+++ b/lib/operations/active_operations.py
@@ -1905,6 +1905,12 @@ class ActiveObjectOperations(BaseObjectOperations) :
                     if _obj_type == "VMC" :
                         _status, _fmsg = _cld_conn.vmcregister(obj_attr_list)
                         _vmcregister = True
+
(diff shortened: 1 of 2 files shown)


hinesmr:
@rayx This commit causes a bug with redis startup. I'm reverting this. Will provide more details.

PR #224, opened 2018-06-14
Jdesfossez/fixes

jdesfossez:
One FIO fix to create the file for randrw workload.
One GCE fix to support setting the size of the root volume.


hinesmr:
Looks great! Thank you!

PR #225, opened 2018-06-14
Revert "Fix a few bugs with gmetad startup code"

hinesmr:
Reverts ibmcb/cbtool#223

@rayx Greetings, my friend. This commit is causing an issue during initial startup of cbtool. What happens is something like this:

```
mariner@mariner-amazon-us-east-1:~/cbtool$ pkill -9 -f redis
mariner@mariner-amazon-us-east-1:~/cbtool$ ./cb -f 
Cbtool version is "8c5fe37"
Parsing "cloud definitions" file..... "/home/mariner/cbtool/lib/auxiliary//../..//configs/mariner_cloud_definitions.txt" opened and parsed successfully.

Checking "Object Store"..... Unable to detect a private Redis server daemon running. Please try to start one (e.g., redis-server /home/mariner/cbtool/lib/auxiliary//../../stores/mariner_redis.conf)
Parsing "cloud definitions" file.....
```

This is very reproducible. So, for the short term, I'm going to revert this. Can you take a look and see what's going on?

What happens above is that the Redis server does in fact get started, but the PR you submitted somehow makes cbtool think that it is dead.

Thank you so much for the contributions, by the way! These PRs have been great........ just a minor bump in the road. :)

### DIFF
--- a/lib/operations/active_operations.py
+++ b/lib/operations/active_operations.py
@@ -1905,12 +1905,6 @@ class ActiveObjectOperations(BaseObjectOperations) :
                     if _obj_type == "VMC" :
                         _status, _fmsg = _cld_conn.vmcregister(obj_attr_list)
                         _vmcregister = True
-
-                        _proc_man = ProcessManagement(username = obj_attr_list["username"], cloud_name = obj_attr_list["cloud_name"])
-                        _gmetad_pid = _proc_man.get_pid_from_cmdline("gmetad.py")
-                        if len(_gmetad_pid) :
-                            cbdebug("Killing the running Host OS performance monitor (gmetad.py)......", True)
-                            _proc_man.kill_process("gmetad.py")
     
                     elif _obj_type == "VM" :
                         self.osci.pending_object_set(_cloud_name, _obj_type, \
@@ -3062,12 +3056,6 @@ class ActiveObjectOperations(BaseObjectOperations) :
                 if _obj_type == "VMC" :
                     self.pre_detach_vmc(obj_attr_list)
                     _status, _msg = _cld_conn.vmcunregister(obj_attr_list)
-
-                    _proc_man = ProcessManagement(username = obj_attr_list["username"], cloud_name = obj_attr_list["cloud_name"])
-                    _gmetad_pid = _proc_man.get_pid_from_cmdline("gmetad.py")
-                    if len(_gmetad_pid) :
-                        cbdebug("Killing the running Host OS performance monitor (gmetad.py)......", True)
-                        _proc_man.kill_process("gmetad.py")
 
                 elif _obj_type == "VM" :
                     self.pre_detach_vm(obj_attr_list)
--- a/lib/remote/process_management.py
+++ b/lib/remote/process_management.py
@@ -34,9 +34,6 @@ class ProcessManagement :
     '''
     TBD
     '''
-
-    allocated_ports = []
-
     @trace
     def __init__(self, hostname = "127.0.0.1", port = "22", username = None, \
                  cloud_name = None, priv_key = None, config_file = None, \
@@ -455,10 +452,9 @@ class ProcessManagement :
         '''
         _port = int(starting_port)
         _pid, _username = self.get_pid_from_port(_port, protocol)
-        while _pid or _port in ProcessManagement.allocated_ports :
+        while _pid :
             _port += 1
             _pid, _username = self.get_pid_from_port(_port, protocol)
-        ProcessManagement.allocated_ports.append(_port)
         return str(_port)
 
     @trace
@@ -536,12 +532,12 @@ class ProcessManagement :
             # Same comment
             sleep(3)
 
-        _pidlist = self.get_pid_from_cmdline(cmdline if not search_keywords else search_keywords)
+        _pid = self.get_pid_from_cmdline(cmdline if not search_keywords else search_keywords)
 
         _msg = "A process with the command line \"" + cmdline + "\" (pid "
-        _msg += str(_pidlist) + ") was started succesfully."
+        _msg += str(_pid) + ") was started succesfully."
         cbdebug(_msg)
-        return _pidlist
+        return [ _pid ]
 
     @trace
     def kill_process(self, cmdline, kill_options = False, port = False) :


hinesmr:
@rayx Merged the revert for now. Please to re-open your PR when you have spare time.

hinesmr:
@rayx Oh, by the way, from what I can tell, the bug seems to only happen when CB is running inside of a Docker container. I'm unable to reproduce it on my laptop, if that helps.

rayx:
@hinesmr I guess it might be an issue with get_pid_from_cmdline() and my code change exposes it. The machine I run CB is shut down today due to power cut. I'll try to reproduce it next week. Thanks.

rayx:
@hinesmr I wonder how you run CB in docker? I tried it and ran into an issue. First I started mongodb and redis manually in docker. Then I started CB and saw this error with "sudo fuser" command:

<pre><code>Checking "Object Store".....An Object Store of the kind "Redis" (shared) on node 10.64.4.5, TCP port 6379, database id "0" seems to be running.
Checking "Log Store".....A Log Store of the kind "rsyslog" (private) on node 10.64.4.5, UDP port 5114 seems to be running.
Checking "Metric Store".....A Metric Store of the kind "MongoDB" (shared) on node 10.64.4.5, TCP port 27017, database id "metrics" seems to be running.
Checking "File Store".....A File Store of the kind "rsync" (private) on node 10.64.4.5, TCP port 873 seems to be running.

Checking for a running API service daemon..... status: Unable to start API and/or GUI daemons: Error while executing the command line "sudo fuser -u 7070/tcp" (returncode = 1) :Cannot stat file /proc/222/fd/0: Permission denied
Cannot stat file /proc/222/fd/1: Permission denied
Cannot stat file /proc/222/fd/2: Permission denied
Cannot stat file /proc/222/fd/3: Permission denied
Cannot stat file /proc/222/fd/4: Permission denied
Cannot stat file /proc/222/fd/6: Permission denied
Cannot stat file /proc/222/fd/7: Permission denied</code></pre>

I ran CB as root so I tried the command manually and reproduce the failure.

<pre><code># fuser -u 7070/tcp  <- no sudo. This was OK

# sudo fuser -u 7070/tcp <- using sudo. It didn't work in container (though it worked in host).</code></pre>

Running CB as a regular user had the same issue. 

I wonder how you get CB running in docker?

rayx:
@hinesmr UPDATE on the "sudo fuser" issue in docker. Since it doesn't relate to CB, I work around it by temporarily removing the 'sudo' in the code. With this change CB starts successfully in docker. Then I apply PR #223. I'm able to run the following commands successfully:

<pre><code># ./cb --hard_reset
# ./cb --soft_reset</code></pre>

Then I try your commands:

<pre><code># pkill -9 -f redis
# ./cb -f          
Cbtool version is "52feed16"
Parsing "cloud definitions" file..... "/cbtool/lib/auxiliary//../..//configs/root_cloud_definitions.txt" opened and parsed successfully.

ERROR:root:network_functions.py/Nethashget.nmap nethashget - Unable to connect to TCP port 6379 on host 10.64.4.5: [Errno 111] Connection refused
Checking "Object Store"..... Unable to detect a shared Redis server daemon running. Please try to start one (e.g., /usr/local/bin/redis-server /etc/redis.conf)
Parsing "cloud definitions" file.....</code></pre>

It does fail as you described. But I don't understand - since the first command kills redis server process, why did you expect the './cf -f' worked? What exactly does the '-f' option do? (I never used it). Does it restart redis? Particularly, since CB uses shared redis by default, I don't think it should try to start redis server in this case, does it?

BTW, I check redis server process with

<pre><code>
# netstat -anp | grep 6379
</code></pre>

There is no redis server process indeed after I kill it with the first command.

rayx:
@hinesmr FWIW, if I start redis server after killing it, cb works.

<pre><code>
# pkill -9 -f redis
# /usr/bin/redis-server --bind 0.0.0.0 &
# ./cb -f
Cbtool version is "52feed16"
Parsing "cloud definitions" file..... "/cbtool/lib/auxiliary//../..//configs/root_cloud_definitions.txt" opened and parsed successfully.

Checking "Object Store".....An Object Store of the kind "Redis" (shared) on node 10.64.4.5, TCP port 6379, database id "0" seems to be running.
Checking "Log Store".....A Log Store of the kind "rsyslog" (private) on node 10.64.4.5, UDP port 5114 seems to be running.
Checking "Metric Store".....A Metric Store of the kind "MongoDB" (shared) on node 10.64.4.5, TCP port 27017, database id "metrics" seems to be running.
Checking "File Store".....A File Store of the kind "rsync" (private) on node 10.64.4.5, TCP port 873 seems to be running.

Executing "soft" reset: (killing all running toolkit processes and flushing stores) before starting the experiment......
    Killing all processes... done
    Flushing Object Store... done
    Flushing Log Store... done


Checking for a running API service daemon.....API Service daemon was successfully started. The process id is 1284 (http://10.64.4.5:7070).
Checking for a running GUI service daemon.....GUI Service daemon was successfully started. The process id is 1307, listening on port 8080. Full url is "http://10.64.4.5:8080".
...
</code></pre>

hinesmr:
You should be running docker with `NET_ADMIN` and PTACE capabilities. That will make your fuser problem go away.

For the pkill solution, yes, that makes it work, but the patch needs to be able to do it without requiring the user to do anything. It should work the same way before and after the patch.

rayx:
@hinesmr Thanks for the suggestion for those capabilities. Didn't know docker requires them.

For the pkill solution, however, I didn't propose it as a solution, but just as an example trying to understand what '-f' did. I did a quick search in the code just now and found it's actually '--soft_reset'. So the problem is that the following command used to work but doesn't with my patch, right?

<pre><code>
# pkill -9 -f redis
# ./cb --soft_reset
</code></pre>

In my CB installation, redis server is set to be "SHARED" by default, so "./cb --soft_reset" should do nothing about it. If so, I don't think the above commands should work even without my patch. Do you set redis server's "SHARED" property to False?

hinesmr:
Yes, that's exactly right. I don't use `shared`, I use `private`.

Can you make the patch work with private?

rayx:
@hinesmr No problem. I'll look into it next week.

rayx:
@hinesmr A quick question. When I set this: 

<pre><code>
[OBJECTSTORE]
USAGE = private
</code></pre>

Do I need to set up other parameters to tell CB how to start and stop redis service? Otherwise how CB knows that?

hinesmr:
@rayx No, no other paramters. CB uses the local IP address to determine how to setup Redis and configures it automatically. You can choose the port or the IP address in that same configuration section as well if you want, but otherwise, it will just use defaults.

PR #226, opened 2018-06-16
A tool to plot graphs for host and VM OS metric data

rayx:
The change introduces a new graph plotting tool. It read host and VM
OS metric data from CSV files generated by monextract command, save
them into RRDtool database and use RRDtool graph command to plot graphs.
The tool is unique in that it doesn't hardcode metrics and graphs, but
provides a config file in YAML format for user to define them. A default
config file is shipped with the tool as an example.

To run the tool, user need to enable host and VM metric collecting and
run monextract to save metric data into CSV files. Then run the tool:

  $ python rrdtool.py <experiment_dir>

The command will use the default config file to generate graphs in the
experiment_dir. For more details about how it works and how to config it,
please read the file header comments in the tool script and the example
config file.


rayx:
This page contains an example output of the command in console:

https://github.com/rayx/rayx.github.io/blob/master/cbtool/rrdplot/console.txt

This page contains example image files generated by the command:

https://github.com/rayx/rayx.github.io/tree/master/cbtool/rrdplot

For example, this image contains all the graphs for a host node:

https://github.com/rayx/rayx.github.io/blob/master/cbtool/rrdplot/EXP-06-05-2018-08-21-38-AM-UTC_host_ss05.png

and this image contains memory graphs of all host nodes and VM nodes (it's experiment containing two AIs):

https://github.com/rayx/rayx.github.io/blob/master/cbtool/rrdplot/EXP-06-05-2018-08-21-38-AM-UTC_memory.png

rayx:
A few more notes:

- I tested the script with a single node cloud, but I think it should work fine with CSV files containing metrics from multiple host nodes because the same code is used to retrieve data from host and VM CSV files.

- I add five new dependencies for the tool and set their tags to 'base'. Please help to change it if it should other value.

maugustosilva:
This is a great addition, thank you very much for your contributions! This tool is superior to the current R scripts that I originally wrote due to its ability to specify which graphics and metrics should be created, right? Definitely merging it, and if it works as intended, I believe we should extend it to plot the management and runtime application metrics.

rayx:
@maugustosilva Glad you like it. The R scripts your wrote are great too, and it have more functions that my tool can provide. One main limitation of my tool is that it's based on RRDtool, which is designed for time series data, so it can't plot graphs for VM provision metrics, or latency vs. load graph, etc. The config file format in its current form probably doesn't apply to those scenarios too.

PR #236, opened 2018-06-25
vmdetach: Retry VM delete requests

hinesmr:
For public clouds, we retry infinitely when it's time to delete a VM.
For everybody else, the default behavior is the same as before.


hinesmr:
@maugustosilva This passes my tests. It doesn't reproduce all the time, because it's difficult to make GCE fail when you want it to, but it passes a SPEC Cloud run.

PR #248, opened 2018-07-27
kernbench workload

jdesfossez:


scripts/kernbench/cb_run_kernbench.sh
```diff
@@ -0,0 +1,60 @@
...
+if test $KERNBENCH_NR_CPUS = 0; then
+	NRJOBS=""
+else
+	NRJOBS="-o $KERNBENCH_NR_CPUS"
+fi
+
+#If more than one command is needed (e.g., connected by "&&" or ";", please dump it on script, instead of just assigining to the variable CMDLINE"
+echo "#!/usr/bin/env bash" > kernbenchloadgen.sh
+echo "cd linux" >> kernbenchloadgen.sh
+echo "../ltp/utils/benchmark/kernbench-0.42/kernbench -n 1 -H -M -f $NRJOBS" >> kernbenchloadgen.sh
```
hinesmr: Here, we want to change this to use ${dir} so that is relative to the cbtool installation directory.

scripts/kernbench/cb_run_kernbench.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Override the default 4*cpu for make -j <nr>
+if test $KERNBENCH_NR_CPUS = 0; then
+	NRJOBS=""
+else
+	NRJOBS="-o $KERNBENCH_NR_CPUS"
+fi
+
+#If more than one command is needed (e.g., connected by "&&" or ";", please dump it on script, instead of just assigining to the variable CMDLINE"
+echo "#!/usr/bin/env bash" > kernbenchloadgen.sh
```
hinesmr: Can this be a static, parameterized script instead of an echo? That way one can repeat the script in isolation to play with things?

scripts/kernbench/dependencies.txt
```diff
@@ -0,0 +1,27 @@
...
+### END - Dependency installation order ###
+
+### START - Dependency-specific installation method ###
+# pm = "package manager" (yum or apt-get)
+# sl = "soft link" (assume that the dependency is already installed, just has to
+# be properly exposed to the user's path.
+# git = git clone using above giturl
+# pip = python pip utility
+# man = "manual"
+kernbench-install = man
```
hinesmr: Man needs to be changed to github so that the tool can automatically install the tool for the user.

hinesmr: With the right URL

scripts/kernbench/virtual_application.txt
```diff
@@ -0,0 +1,40 @@
...
+PROFILES = default
+REFERENCE = https://github.com/linux-test-project/ltp
+LICENSE = GPL_v2
+REPORTED_METRICS = elapsed_time, user_time, system_time, percent_cpu, context_switches, sleeps
+
+# VApp-specific MANDATORY attributes
+DESCRIPTION =Deploys a single instance and runs the "kernbench" synthetic benchmark\n
+DESCRIPTION +=  - LOAD_PROFILE possible values: _PROFILES_.\n
+DESCRIPTION +=  - LOAD_LEVEL meaning: number of jobs running in parallel, currently only 1 is supported.\n
+DESCRIPTION +=  - LOAD_DURATION meaning: not used, a run ends when a specific\n
```
hinesmr: We need the block / remote storage configuration options here and for the script to support using them and placing the kernbench files into block storage, should the user request it.

scripts/kernbench/virtual_application.txt
```diff
@@ -0,0 +1,40 @@
...
+DESCRIPTION +=    number of I/O operations is reached.\n
+KERNBENCH_SETUP1 = cb_start_kernbench.sh
+START = cb_run_kernbench.sh
+
+# VApp-specific modifier parameters. Commented attributes imply default values assumed
+KERNBENCH_NR_CPUS = 0
```
hinesmr: What does a `0` default do?

PR #252, opened 2018-10-11
DO k8s: First working implementation

hinesmr:
This commit does a few things:

1. We inherit from the base class in kub_cloud_ops.py and leave the file pretty much unchanged
2. We then create DigitalOcean clusters on demand. (There is no libcloud support to do this, unfortunately)
3. Then we're able to use all of the existing k8s functions to create containers with very little changes.

It works! It's probably got bugs, but it works.

Because it's k8s, it has a really nice side effect that once the cluster is created, all future commands
go to k8s, so in terms of error handling and things like that, all future code can remain
in the original "kub" adapter without having to extend too much error handling to all of the
other cloud providers.

We should be able to use the file "dok8s_cloud_ops.py" as a pretty good template for the remaining clouds.

### DIFF
--- a/configs/templates/_dok8s.txt
+++ b/configs/templates/_dok8s.txt
@@ -1,0 +1,163 @@
+[USER-DEFINED]
+DOK8S_ACCESS = https://api.digitalocean.com/v2 
+# Name of the ssh key pair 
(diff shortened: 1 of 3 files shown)


hinesmr:
This is not ready to merged yet, just here for discussion.

PR #256, opened 2018-12-10
API: add state to 'list_objects' command

hinesmr:
This allows the `applist` command to return the state value as a key/value in the returned dictionary output. Currently, it's not there.


jdesfossez:
LGTM

PR #257, opened 2018-12-14
Nested containers support

mraygalaxy:
This change is ready for review and allows us to pull the docker container images directly into a completely empty base virtual machine (say, Ubuntu 18.04, for example) with absolutely zero dependencies on anything whatsoever.

Features:

1. It supports both public and private docker registries, by default using the public `ibmcb` group of containers.
2. Private registries can be both HTTPS or insecure.
2. It supports the VPN.
3. It supports external volumes mounted within the nested container.
4. If you "pre-cache" a base image with all of the docker container images, it will definitely speed up bootup time and still get the same benefits. Any changes to the container will get pulled the next time the base image boots without having to ever regenerate a new snapshot.

It's very configurable, here's an example:

> [VM_DEFAULTS]
> NEST_CONTAINERS_ENABLED = $True
> NEST_CONTAINERS_REPOSITORY = 10.9.0.1:5000
> NEST_CONTAINERS_INSECURE_REGISTRY = $True

By default the registry above is `ibmcb`, if you don't override it.

> [VM_TEMPLATES]
> NEST_CONTAINERS_BASE_IMAGE = size:NA, imageids:1, imageid1:18.04_x64, cloudinit_packages:openvpn;bc;jq;docker.io;python;redis-server

Note the `18.04_x64` above.

Next, we introduce a new object in CloudBench:

> [CONTAINER_TEMPLATES]
TINYVM = imageids:1, imageid1:ubuntu_cb_nullworkload

All of this put together works by `transparently` replacing the Image (tinyvm, for example) in the `VM_TEMPLATES` section with "common" specified by  NEST_CONTAINERS_BASE_IMAGE. The user does not need to modify that section, but simply populate a new section that specifies with container images matchup with the AI / VM types.

Note the additional requirements for `cloudinit_packages`, such as python and redis and docker, which don't typically come pre-installed with a completely empty virtual machine. Cloud-init is thus mandatory for this all to work.

### DIFF
--- a/configs/templates/internal_options.txt
+++ b/configs/templates/internal_options.txt
@@ -536,3 +536,4 @@ SYSBENCH = imageids:1, imageid1:ubuntu_cb_sysbench
 MYSQL = imageids:1, imageid1:ubuntu_cb_sysbench
 BONNIE = imageids:1, imageid1:ubuntu_cb_bonnie
 LINPACK = imageids:1, imageid1:ubuntu_cb_linpack
+KERNBENCH = imageids:1, imageid1:ubuntu_cb_kernbench


hinesmr:
This is ready for review now.

PR #258, opened 2018-12-21
libcloud: ability to log http headers

hinesmr:
We need the ability to log HTTP headers to debug what happens
with libcloud, but libcloud doesn't make that easy, so we have to
bypass it.

What we're trying to do here is solve a conundrum. Under the covers,
libcloud uses /usr/lib/python2.7/httplib.py as it's primary HTTP
transport library.

httplib, has the ability to debug itself, but instead of using the logging
library like a "good" python library should, it just prints to stdout.
In a separate patch (via Dockerfile) we fix httplib to work correctly
and use a logger.

That being said, after it uses the logging library correctly, we still
have to capture the log output the headers ONLY when something bad happens.
We don't want to spew a bunch of header logs to the logging system
for the 99% of the time when things are working fine.

To solve that problem, we trap the (newly fixed) http header log messages
from the httplib library into a StringIO buffer. Then, if we happen
to hit a python exception in libcloud, we dump those headers to the log
if and only if there is an exception.

### DIFF
--- a/lib/clouds/libcloud_common.py
+++ b/lib/clouds/libcloud_common.py
@@ -32,14 +32,68 @@ from libcloud.compute.types import NodeState
 
 from copy import deepcopy
 
+from cStringIO import StringIO
+
 import threading
 import traceback
+import os
+
+try:
+    from http.client import HTTPConnection # py3
+    from http.client import HTTPSConnection # py3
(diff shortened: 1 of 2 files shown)


hinesmr:
This is ready for review now.

PR #260, opened 2019-01-10
kernbench workload

jdesfossez:
Add the kernbench workload, each run cleans up the tree, make a
defconfig and then compile the kernel with the specified amount of
compiler threads (default is 4 * cpus).


hinesmr:
Looks great!

PR #263, opened 2019-02-06
libcloud: Don't delete too many times

hinesmr:
For whatever reason, there were one to may destroy() operations being submitted. Let's fix that.


jdesfossez:
Tested and confirm it prevents the loop of repeated destroy events

PR #270, opened 2019-02-22
k8s Fix for private registries and VPN connectivity

hinesmr:
1. The registry was crashing due to an obscure bug upstream when running inside the cluster [fixed]
2. We needed to define an additional service for VPN to work properly instead of using the Pod IP address (which can change) [fixed]


jdesfossez:
LGTM

PR #272, opened 2019-02-26
Improvements and bugfixes for instance configuration auxiliary functions

maugustosilva:
1) Given that some workloads (e.g., Hadoop) make use of bashrc,
cb_post_boot now ensures that ~/.bashrc always exists during
initial execution.
2) Make sure that function `check_container` always returns the
correct number of cpus and and amount of memory
3) Make sure that function `mount_filesystem_on_volume` will not
accidentally "double mount" filesystem, whether executed on containers
or on VMs/Bare-Metal.
4) Add a new function, `fix_ulimit` which prevents application-level
errors due to open file description exhaustion (among others)
5) Make sure that directories are automounted as early as possible
(the function `automount_data_dirs` and associated functions are
idempotent, so it shouldn't be a problem calling it during the
generic startup, and let developers to call it again during
application-specific startup configuration).

### DIFF
--- a/scripts/common/cb_common.sh
+++ b/scripts/common/cb_common.sh
@@ -73,6 +73,16 @@ fi
 export PATH=$PATH:~
 eval PATH=$PATH
 
+if [[ ! -f ~/.bashrc ]]
+then
+    sudo ls /root/.bashrc > /dev/null 2>&1
+    if [[ $? -eq 0 ]]
+    then
+        sudo cp /root/.bashrc ~/.bashrc
+        sudo chown $(whoami):$(whoami) ~/.bashrc
+    fi
+fi  
+
 # Test for redis-cli
 if [ x"$(uname -a | grep CYGWIN)" != x ] ; then
     rediscli="$dir/redis.bat"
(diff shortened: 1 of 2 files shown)


scripts/common/cb_common.sh
```diff
@@ -593,7 +611,14 @@ function mount_filesystem_on_volume {
                 exit 1
             else
                 syslog_netcat "$FILESYS_TYPE filesystem on volume $VOLUME successfully made accessible through mountpoint ${MOUNTPOINT_DIR}"
+                
+                if [[ $(sudo cat /etc/fstab | grep -c ${MOUNTPOINT_DIR}) -eq 0 ]]
+                then
+                    sudo bash -c "echo \"${VOLUME} ${MOUNTPOINT_DIR} ext4  defaults 0 2\" >> /etc/fstab"
```
hinesmr: Should `ext4` be hard-coded here? Is it not the same as `$FILESYS_TYPE` ?

maugustosilva: Of course! My mistake. Fixed and pushed.

PR #281, opened 2019-02-26
Fixes for Sysbench

maugustosilva:


scripts/sysbench/cb_restart_mysql.sh
```diff
@@ -25,7 +25,7 @@ MYSQL_NONROOT_USER=`get_my_ai_attribute_with_default mysql_nonroot_user sysbench
 MYSQL_NONROOT_PASSWORD=`get_my_ai_attribute_with_default mysql_nonroot_password sysbench`
 MYSQL_DATA_DIR=`get_my_ai_attribute_with_default mysql_data_dir /sysbench`
 MYSQL_CONF_FILE=`get_my_ai_attribute_with_default mysql_conf_file /etc/mysql/mysql.conf.d/mysqld.cnf`
-MYSQL_RAM_PERCENTAGE=`get_my_ai_attribute_with_default mysql_ram_percentage 70`
+#MYSQL_RAM_PERCENTAGE=`get_my_ai_attribute_with_default mysql_ram_percentage 10`
```
hinesmr: Why are we removing this functionality?

maugustosilva: We shouldn't :-) Re-added it, properly taking into account when we are running on containers.

PR #313, opened 2019-08-20
A significant overhauling of the KVM/QEMU image builder for Cloudbench

maugustosilva:
1) Check all dependencies in "green field" node before even attempting
to build (should make significantly easier for users to create images
"out of the box")
2) The (required) rsync server (used, for instance for the building of
"coremark" Application Instance type) is started as part of the building
process
3) Make sure the "machine-id" on the image is cleaned up at the end of
the building process, ensure that every VM deployed by CB will have an
unique image ID (needed, for instance, for the correct IP assignment by
libvirt's DHCP server)


hinesmr:
@maugustosilva Looks like you might need to rebase ^^^

PR #316, opened 2019-08-21
speed up multi-region cloud images

hinesmr:
For GCE and DigitalOcean, we have this odd behavior with the way images are managed. We use "global" images and snapshots => meaning that images have the same ID / name regardless of which region they execute. So, if you use multiple VMCs (which is the case with google and digitalocean) at the same time (such as an availability zone), then you end up polling the API for the list of images multiple times for no reason. The additional polls for the other regions end up being duplicate polls and are not actually required and take a lot of time if you're doing complex tests. Amazon doesn't have this behavior...... each "AMI" is unique for each region, so it doesn't apply.

This change makes a simple modification to libcloud in such a way that we only need to poll for the imagelist one time at startup and makes things go faster.


jdesfossez:
LGTM

PR #317, opened 2019-08-26
Improvements and bugfixes in common bash functions

maugustosilva:
- When an user specifies, in the AI-specific <ROLE>_DATA_DIR attribute
the value "none" (indicating that a given application will use the volume without a filesystem on it), the function mount_filesystem_on_volume can now be called without resulting in error
- The function check_container now correctly detects the number of CPUs and amount of memory available, in KB.
- Make sure every other function (e.g., haproxy_setup) uses the function check_container

### DIFF
--- a/scripts/common/cb_common.py
+++ b/scripts/common/cb_common.py
@@ -546,7 +546,10 @@ def is_number(val) :
     '''
     try:
         _val = float(val)
-        return _val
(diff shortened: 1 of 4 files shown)


hinesmr:
I'll test this one in a little bit with the wrk2 workload to make sure everything's still running on my end, one moment.........

PR #345, opened 2019-10-16
Add ability to automatically configure firewall on instances

maugustosilva:
By setting the attribute `CONFIGURE_FIREWALL` to `$True`, a new
`configure_firewall` function will be run during the "post-boot" phase
will make sure that instances can be accessed only by the Orchestrator
node and instances that are part of the same AI. The main use case for
it is public clouds where instances get a directly accessible public IP.
While an experimenter can (and should) configure the cloud-specific
controls for "virtual networking security groups", at an small cost in
terms of code, we can provide a cloud-agnostic solution that should
protect everything out of the box.

### DIFF
--- a/configs/templates/internal_options.txt
+++ b/configs/templates/internal_options.txt
@@ -224,6 +224,7 @@ PREPARE_IMAGE_NAME = None
 NEST_CONTAINERS_ENABLED = $False
 NEST_CONTAINERS_REPOSITORY = ibmcb
 NEST_CONTAINERS_INSECURE_REGISTRY = $False
+CONFIGURE_FIREWALL = $False
 # Use the 'openssl' command to choose a password to enable this feature.
 # Requires cloud-init support.
 PASSWORD = $False # Disabled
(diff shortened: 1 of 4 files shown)


configs/templates/internal_options.txt
```diff
@@ -224,6 +224,7 @@ PREPARE_IMAGE_NAME = None
 NEST_CONTAINERS_ENABLED = $False
 NEST_CONTAINERS_REPOSITORY = ibmcb
 NEST_CONTAINERS_INSECURE_REGISTRY = $False
+CONFIGURE_FIREWALL = $False
```
hinesmr: Could we make the true by default? I kind of want to be forced to experience the pain while we practice packaging CloudBench......... by making this true by default, it will also help others gain more confidence and trust (including SPEC).

maugustosilva: Very good point... will do.

hinesmr:
@maugustosilva This also needs to be rebased with master before I can test it.

PR #356, opened 2019-11-27
Dealing with multi-vnic instances

maugustosilva:
This commit adds the ability to deal with multi-vnic instances. Not
only CB will automatically detect and run dhclient on multiple
additional vNICs, it will also collect all the IP addresses and store
those on the VM-specific variable "extra_cloud_ips".
Also, it is included a better, more robust function to set the mtu on
vNICs, which shall be used in future updates.

### DIFF
--- a/scripts/common/cb_common.sh
+++ b/scripts/common/cb_common.sh
@@ -742,7 +742,7 @@ function mount_remote_filesystem {
 }
 export -f mount_remote_filesystem
 
-my_if=$(netstat -rn | grep UG | awk '{ print $8 }')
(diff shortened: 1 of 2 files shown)


scripts/common/cb_common.sh
```diff
@@ -1150,7 +1149,15 @@ function refresh_hosts_file {
     fi
 
     syslog_netcat "Refreshing hosts file ... "
-    sudo bash -c "rm -f /etc/hosts; echo '127.0.0.1    localhost' >> /etc/hosts; cat ${ai_mapping_file} >> /etc/hosts"
```
hinesmr: This appears to duplicate work from a previous PR?

PR #374, opened 2020-02-07
Add missing kernbench config entry

el8:
Kernbench was only added to container templates so far.
Add a kernbench entry to vm_templates to make it runnable
inside a VM.

Signed-off-by: Jan Glauber <jglauber@digitalocean.com>


jdesfossez:
LGTM

PR #384, opened 2020-05-26
Fix python3 syntax

jdesfossez:
Only fixed the syntax so the cbtool module installed from pip can run
with python3. Everything still runs fine with Python 2.7.

After making the change and ensuring we can import the pip package in python3, I tested various CB functions (attaching to VMC, creating/stopping AIs, etc) with python2.7 and everything is still working as is.

The changes I made are compatible with python3 and python2.7 and I included a couple of version checks at the places where a breaking code change was required

### DIFF
--- a/lib/api/api_service_client.py
+++ b/lib/api/api_service_client.py
@@ -32,12 +32,17 @@
 '''
 
 from sys import path
-from xmlrpclib import Server
-import xmlrpclib
 import pwd
 import sys
 import re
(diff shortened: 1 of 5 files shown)


mraygalaxy:
I'm almost done with the port in #385 . That will supercede that, so I'll go ahead and close this. The bindings have already been pushed to PyPi as version 3.0.1

PR #387, opened 2020-06-22
Remove dynamic API binding logic

mraygalaxy:
This code is a relic from another era. We used to run multiple cloud
attachments in the same CB instance, which is still supported, but
not while the VPN is running. Just don't do that anymore.


jdesfossez:
LGTM

PR #393, opened 2020-07-22
libcloud: Match cloud names correctly upon deletion.

mraygalaxy:
We were previously (greedily) using count() instead of matching exactly.


jdesfossez:
LGTM

PR #394, opened 2020-07-23
Redis: flush only exactly the cloud that was requested.

mraygalaxy:
When clouds share substrings in common, the wrong keys end up getting deleted.


jdesfossez:
LGTM

PR #398, opened 2020-08-26
Fix in GUI and haproxy config

jdesfossez:


scripts/common/cb_common.sh
```diff
@@ -2196,7 +2196,7 @@ EOF
     do
         for BACKEND_IP in $LOAD_BALANCER_BACKEND_SERVERS
         do
-            echo "  server $(cat /etc/hosts | grep $BACKEND_IP | grep -v lost | awk '{ print $2 }') $BACKEND_IP:$LBP check" >> $f
+            echo "  server $(cat /etc/hosts | grep $BACKEND_IP | grep -v lost | head -1 | awk '{ print $2 }') $BACKEND_IP:$LBP check" >> $f
```
mraygalaxy, line 2199: @jdesfossez Can you talk about this one? I can't remember why I added the head -1, but what was the before/after?

jdesfossez: it's an issue I just found, there are duplicates in /etc/hosts, so what happens by default is something like this:
```
$ echo "  server $(cat /etc/hosts | grep 10.39.32.68 | grep -v lost | awk '{ print 
$2 }') $BACKEND_IP:$LBP check"
  server 10.39.32.68   cb-mariner-DROPLETS2R5JDCORESCHEDNOISE-vm174
10.39.32.68    cb-mariner-droplets2r5jdcoreschednoise-vm174    #tinyvm        ,0D2D6523-5B77-51C6-A178-CD014F85463D
10.39.32.68    cb-mariner-DROPLETS2R5JDCORESCHEDNOISE-vm174 : check
```
which is broken, so by limit to 1 entry I get a valid line:
```
$ echo "  server $(cat /etc/hosts | grep 10.39.32.68 | grep -v lost | head -1 | awk '{ print 
$2 }') $BACKEND_IP:$LBP check"
  server 10.39.32.68   cb-mariner-DROPLETS2R5JDCORESCHEDNOISE-vm174 : check
```

mraygalaxy: Perfect, thank you :)

mraygalaxy:
LGTM

PR #402, opened 2020-10-21
Support additional FIO config options within cbtool

vermavis:
This adds support for some further FIO config options
within cbtool. Following FIO config params were added:

       - thinktime
       - thinktime_blocks
       - rate_process
       - bssplit

The default behavior is to not use them unless someone enable
them in the FIO based AI config.


mraygalaxy:
Looks great. I love it when options are added that I've never seen before. Thanks, Vishal!

PR #406, opened 2020-11-13
fix MetricStoreMgdConnException issue

vermavis:
We faced an issue within cb in which when the code path
was touching this user-defined exception "MetricStoreMgdConnException"
python3 complained with following msg:

"TypeError: catching classes that do not inherit from BaseException
is not allowed."

Due to this, the client API object was getting hung and the client API kept
waiting to receive something from the socket. This fixes the issue
by using local version of MetricStoreMgdConnException class and
because of that we no longer see that TypeError msg.


maugustosilva:
This one looks good, and fixes a bug I was seeing too. Merged

PR #407, opened 2020-11-19
Set of minimal changes to ensure a complete successful regression test (py3)

maugustosilva:
This PR contains all changes required for a successful simulated
regression test, as well as a minimal (basic VM creation) for all
supported clouds.

The problem with daemonization seems to be satisfactorily solved with
python-daemon-3K


jdesfossez (approved):
That looks like a nice cleanup, LGTM

PR #408, opened 2020-12-02
update Linpack workload to support AMD systems

vermavis:
Cb currently uses an Intel optimized linpack version which runs only
on Intel based systems. This updates Linpack workload to version 2.3
and it also adds support for running Linpack workload on AMD based
systems along with Intel based systems.  A new config option:
"linpack_version" is added which by default will just run the current
linpack version. However, if the user sets it to 2.3 then the updated
linpack version will run.

### DIFF
--- a/scripts/linpack/cb_run_linpack.sh
+++ b/scripts/linpack/cb_run_linpack.sh
@@ -22,47 +22,161 @@ source $(echo $0 | sed -e "s/\(.*\/\)*.*/\1.\//g")/cb_common.sh
 set_load_gen $@
 
 LOAD_PROFILE=$(echo ${LOAD_PROFILE} | tr '[:upper:]' '[:lower:]')
+LINPACK_VERSION=`get_my_ai_attribute_with_default linpack_version 2.3`
 
-LINPACK=`get_my_ai_attribute_with_default linpack ~/compilers_and_libraries_2016.0.038/linux/mkl/benchmarks/linpack/xlinpack_xeon64`
-eval LINPACK=${LINPACK}
+# Use version 2.3 if specified
+if [[ $LINPACK_VERSION == "2.3" ]]; then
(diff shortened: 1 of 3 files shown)


mraygalaxy:
@vermavis I don't see any problem with just defaulting it to the latest 2.3 version. Would that help you?

scripts/linpack/cb_run_linpack.sh
```diff
@@ -22,47 +22,161 @@ source $(echo $0 | sed -e "s/\(.*\/\)*.*/\1.\//g")/cb_common.sh
...
+	execute_load_generator "$CMDLINE" ${RUN_OUTPUT_FILE} ${LOAD_DURATION}
+	RESULTS=$(cat ${RUN_OUTPUT_FILE} | grep -A 1 Average | grep $PROBLEM_SIZES)
+	AVERAGE=$(echo $RESULTS | awk '{print $4}')
+	MAX=$(echo $RESULTS | awk '{print $5}')
     
+	~/cb_report_app_metrics.py \
+	throughput_max:$MAX:gflops \
+	throughput:$AVERAGE:gflops \
+	$(common_metrics)
+fi
```
mraygalaxy, line 179: Could we just make the older version obsolete and remove the 2nd part of this?

vermavis: Yes thats actually easier. I was not sure for the case when someone would like to compare previous linpack run results? Maybe mark the new one default and have a way to run the older version if someone wants to use it? 

I am OK either way..

vermavis: Just updated it so 2.3 becomes the default version. Still keeping the older version so we have a way to transition between older and newer results. Let me know what you think.

mraygalaxy: Looks great. I can see how radically different the two version in. Very nice patch. Thank you sir.

PR #409, opened 2020-12-07
fix MySQL adapter logging

vermavis:
We noticed an issue that the log messages from MySQL adapter class
were not being written to the operations log file. This fixes it.


maugustosilva:
hello @vermavis  is this PR still relevant? https://github.com/ibmcb/cbtool/pull/409

PR #410, opened 2020-12-09
Fix monextract

jdesfossez:
There was the same issue about the MySQL exception handler that was
fixed in 18d164d685af1a9b5de3e3c100e87b85c6b6359f

Then some typos (obj_attr_list -> _obj_attr_list).

And finally, some files need to be open with buffering.

### DIFF
--- a/configs/templates/_do.txt
+++ b/configs/templates/_do.txt
@@ -84,7 +84,7 @@ SECURITY_GROUPS = not_yet_applicable
 GLOBAL_IMAGES = $True
 
 [VM_TEMPLATES : DO_CLOUDCONFIG ]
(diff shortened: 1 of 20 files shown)


maugustosilva:
Hi @jdesfossez. I had this fixed on PR #407 (it was required, otherwise a complete regression test wouldn't pass). Can you take a look that PR and tell me if you're ok with merging it? Many thanks in advance!

jdesfossez:
Oh, right, I missed that !
I will review and update/close this PR accordingly.

jdesfossez:
Ok, I updated this one with only the 2 small fixes missing from your branch.

PR #422, opened 2022-01-14
Fix dependencies Checker

jpavlav:
Hello @mraygalaxy! I work with @jdesfossez at DigitalOcean. During some testing of `cbtool` on Python3.8, ran into an issue with the dependencies checker. As it turns out the `linux_distribution()` method was removed from the `platform` module in the standard lib in Python3.7.

Related Python Bugs (discussion about removing the method):

* https://bugs.python.org/issue1322
* https://bugs.python.org/issue28167

* Try to import `distro`. If version below 3.7, pass, else, raise
exception. `distro` is a requirement for Python >= 3.7
* Try to use `platform.linux_distribution()`. If an `AttributeError` is
raised (indicating the `linux_distribution` method is no longer
available), use `distro.name()`, `distro.version()`, `distro.id()` to
obtain the same information from the `distro` package.
* Include `distro` in `setup.py` to cover package installation.

### DIFF
--- a/lib/auxiliary/dependencies.py
+++ b/lib/auxiliary/dependencies.py
@@ -48,7 +48,7 @@ except ModuleNotFoundError as e:
             "Version: %s" % platform.python_version()
         )
         cberr(_msg)
-        raise Exception(e)
+        raise e
     pass
 
 def deps_file_parser(depsdict, username, options, hostname, process_manager = False) :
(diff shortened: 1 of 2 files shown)


mraygalaxy:
@jpavlav Hi Justin, very nice to meet you. I'll go ahead and merge this one now.

Have a good weekend. = )

jpavlav:
@jpavlav Thank you for the quick merge! Curious, do you think that this latest master can be pushed up to PyPI?

mraygalaxy:
@jpavlav  Done! I just uploaded version 3.0.5.

PR #423, opened 2022-01-27
coremark: fix build parameters

jdesfossez:


jdesfossez:
something is not right, it seems to only capture results with `-DVALIDATION_RUN=1`, I need to read more, but I don't think we are supposed to start it like that, just confused at their doc :/

PR #429, opened 2023-08-17
Linode followup bugfixes

mraygalaxy:
1. Linode doesn't have the same power-on command as other clouds, so handle that better. We have a policy in cloudbench that we do not destroy offline instances (for evidence purposes, in case a VM or host was broken and needs to be investigated), so this is an important command.

2. I want to revert this commit from 2018:

commit 751de25fba78c99ed4db0b24d146a48fc9371a98
Date:   Fri Mar 2 10:12:37 2018 -0500

If we keep this commit, it causes our files to be owned by root inside the running containers instead of being owned by 'cbuser' and we're not able to startup successfully.

### DIFF
--- a/lib/clouds/libcloud_common.py
+++ b/lib/clouds/libcloud_common.py
@@ -605,6 +605,15 @@ class LibcloudCmds(CommonCloudFunctions) :
         return [ ]
 
     @trace
+    def power_on_node(self, credentials_list, instance) :
+        adapter = self.get_adapter(credentials_list)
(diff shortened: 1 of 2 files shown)


maugustosilva:
We indeed should revert that particular commit, it was an ill-advised and poorly thought-through attempt at solving a temporary problem which caused far reaching consequences down the road.

maugustosilva (approved):

mraygalaxy:
Thank you sir. = )

PR #431, opened 2023-09-05
Update workloads to 22.04

mraygalaxy:
Not all workloads are 22.04 (Jammy) compatible:

1. sysbench-tpcc support stopped at 20.04.
2. mongodb was removed as a native Ubuntu package and requires using the company's PPA
3. We are still pinning cassandra for SPEC Cloud purposes. Is that still necessary given the current state of SPEC Cloud development?
4. The linpack URL changed yet again.
5. python2.7 was still being used in some places. We need to finish deprecating that.
6. Add a workload dockerfile for stress-ng. (I thought I did that already, but I guess not)

### DIFF
--- a/docker/build_common.sh
+++ b/docker/build_common.sh
@@ -61,7 +61,7 @@ else
     echo "##### rsync server was found on ${CB_RSYNC}"  
 fi
 
-CB_UBUNTU_BASE=ubuntu:18.04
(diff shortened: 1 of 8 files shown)


maugustosilva (approved):

maugustosilva:
Very welcome fix. For SPECCloud, we just need a minimal version of Cassandra.

PR #434, opened 2023-12-11
convert the byte arrays to strings using UTF-8 encoding

tli16:
In recent tests we noticed that the tables names are being returned as byte arrays in python3 while `_latest_tables` and `_indexed_tables` contains regular strings, which leads to the failure of finding table correctly. 
For instance, `latest_management_VM_username` is shown as `bytearray(b'latest_management_VM_username')` in `_tables_found`. 

Convert the byte arrays to regular strings using UTF-8.

### DIFF
--- a/lib/stores/mysql_datastore_adapter.py
+++ b/lib/stores/mysql_datastore_adapter.py
@@ -163,7 +163,7 @@ class MysqlMgdConn(MetricStoreMgdConn) :
             _tables_found = [] 
 
             for x in cursor:
-              _tables_found.append(x[0])
+              _tables_found.append(x[0].decode('utf-8'))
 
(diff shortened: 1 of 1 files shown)


maugustosilva:
@mraygalaxy @mraygalaxy2 I believe you are the one using the MySQL backend more heavily. Are you ok with this merge?

mraygalaxy:
LGTM. This is just in the "intialize" routine, right, so it doesn't affect the main data path. Thanks for the heads up, Marcio.

mraygalaxy (approved):
LGTM. This is just in the "intialize" routine, right, so it doesn't affect the main data path. Thanks for the heads up, Marcio.

